PR #17, opened 2024-05-31
feat: Add order status helper method

yusufttur:
Add order status helper method with pre-defined statuses.

### DIFF
--- a/src/services/ShopwareDataHelpers.ts
+++ b/src/services/ShopwareDataHelpers.ts
@@ -234,3 +234,8 @@ export function extractIdFromUrl(url: string): string | null {
     const segments = url.split('/');
     return segments.length > 0 ? segments[segments.length - 1] : null;
 }
+
+type OrderStatus = 'cancel' | 'complete' | 'reopen' | 'process';
+export const setOrderStatus = async (orderId: string, orderStatus: OrderStatus , adminApiContext: AdminApiContext): Promise<void> => {
+    await adminApiContext.post(`./_action/order/${orderId}/state/${orderStatus}`);
+};
--- a/src/services/ShopwareDataHelpers.ts
+++ b/src/services/ShopwareDataHelpers.ts
@@ -1,3 +1,4 @@
+import { APIResponse } from '@playwright/test';
 import { AdminApiContext } from './AdminApiContext';
 
 export const getLanguageData = async (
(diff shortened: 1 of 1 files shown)


src/services/ShopwareDataHelpers.ts
```diff
@@ -233,4 +233,9 @@ export const getMediaId = async (fileName: string, adminApiContext: AdminApiCont
 export function extractIdFromUrl(url: string): string | null {
     const segments = url.split('/');
     return segments.length > 0 ? segments[segments.length - 1] : null;
-}
\ No newline at end of file
+}
+
+type OrderStatus = 'cancel' | 'complete' | 'reopen' | 'process';
+export const setOrderStatus = async (orderId: string, orderStatus: OrderStatus , adminApiContext: AdminApiContext): Promise<void> => {
+    await adminApiContext.post(`./_action/order/${orderId}/state/${orderStatus}`);
```
Phil23: Add a return of the API call, so the external user could react to the result of the call.

Phil23 (changes requested):

src/services/ShopwareDataHelpers.ts
yusufttur: Should I add all the data, including the response from the call, or just the transaction ID?
This is how it looks like:
{
    "extensions": {
        "foreignKeys": {
            "extensions": [],
            "apiAlias": "state_machine_state_foreign_keys_extension"
        },
        "internal_mapping_storage": {
            "extensions": [],
            "apiAlias": null
        }
    },
    "_uniqueIdentifier": "018fe21a11e6718a83122821830392b0",
    "versionId": null,
    "translated": {
        "name": "Cancelled",
        "customFields": []
    },
    "createdAt": "2024-06-04T07:15:10.441+00:00",
    "updatedAt": null,
    "name": "Cancelled",
    "technicalName": "cancelled",
    "stateMachineId": "018fe21a11e6718a8312282180e1dfd2",
    "stateMachine": null,
    "fromStateMachineTransitions": null,
    "toStateMachineTransitions": null,
    "translations": null,
    "orders": null,
    "orderTransactionCaptures": null,
    "orderTransactionCaptureRefunds": null,
    "orderTransactions": null,
    "orderDeliveries": null,
    "fromStateMachineHistoryEntries": null,
    "toStateMachineHistoryEntries": null,
    "customFields": null,
    "id": "018fe21a11e6718a83122821830392b0"
}

yusufttur: In other words, is this change good?
```
type OrderStatus = 'cancel' | 'complete' | 'reopen' | 'process';
export const setOrderStatus = async (orderId: string, orderStatus: OrderStatus , adminApiContext: AdminApiContext): Promise<string> => {
    const resp = await adminApiContext.post(`./_action/order/${orderId}/state/${orderStatus}`);
    const result = (await resp.json()) as { data: { id: string }; };
    return result.data.id;
};
```

yusufttur: Return added to API call

Phil23 (approved):

PR #18, opened 2024-05-31
feat: Add flow builder detail page object with allow args in goto metod

yusufttur:


src/page-objects/AdministrationPages.ts
```diff
@@ -51,4 +54,8 @@ export const test = base.extend<FixtureTypes>({
     AdminFlowBuilderListing: async ({ AdminPage }, use) => {
         await use(new FlowBuilderListing(AdminPage));
     },
+    AdminFlowBuilderDetail: async ({ AdminPage }, use) => {
```
Phil23, line 58: Missing white space

src/types/PageObject.ts
```diff
@@ -3,5 +3,5 @@ import type { Page } from '@playwright/test';
 export interface PageObject {
     readonly page: Page;
 
-    goTo(): Promise<void>;
+    goTo(...args: any): Promise<void>;
```
Phil23, line 7: You have to add the ignore of the eslint rule, if you want to use the `any` type here, but maybe there is a better solution.

This also isn't enough, because it will break the implementation of the actor `goesTo` method. We have to implement this as a general solution.

Phil23 (changes requested):

Phil23:
We will provide a general refactoring of the page object goTo method.

PR #32, opened 2024-06-10
Refactor page object usage

Phil23:
To remove the hard dependency between data fixtures and page objects, I removed the usage of the page objects `goTo` method and refactored the way the `goesTo()` method of the actor works.

This has the following advantages:
- No dependency between data fixtures and page objects
- The `goesTo()` method of the actor now uses a url instead of a page object which makes the usage more flexible.
- The page objects have a new parameterized `url()` method which can be used in combination with the changed `goesTo()` method.

Before:
```
await ShopCustomer.goesTo(StorefrontProductDetail);
```

After:
```
await ShopCustomer.goesTo(StorefrontProductDetail.url(ProductData));
```

I also added a test for the general usage of all page objects.

### DIFF
--- a/src/page-objects/administration/CustomerDetail.ts
+++ b/src/page-objects/administration/CustomerDetail.ts
@@ -12,7 +12,7 @@ export class CustomerDetail implements PageObject {
         this.accountCard = page.locator('.sw-customer-card')
     }
 
-    async goTo({ customerId }: Record<string, unknown>) {
-        await this.page.goto(`#/sw/customer/detail/${customerId}/base`);
+    url(customerId: string | undefined) {
+        return `#/sw/customer/detail/${customerId}/base`
     }
 }
--- a/src/page-objects/administration/CustomerDetail.ts
+++ b/src/page-objects/administration/CustomerDetail.ts
@@ -12,7 +12,7 @@ export class CustomerDetail implements PageObject {
         this.accountCard = page.locator('.sw-customer-card')
     }
 
-    url(customerId: string | undefined) {
+    url(customerId: string) {
         return `#/sw/customer/detail/${customerId}/base`
     }
 }
--- a/src/page-objects/administration/CustomerDetail.ts
+++ b/src/page-objects/administration/CustomerDetail.ts
@@ -1,24 +1,18 @@
 import type { Page, Locator } from '@playwright/test';
 import type { PageObject } from '../../types/PageObject';
-import type { components } from '@shopware/api-client/admin-api-types';
 
 export class CustomerDetail implements PageObject {
-    public readonly customerId;
-    public readonly customer: components['schemas']['Customer'];
-
     public readonly editButton: Locator;
     public readonly generalTab: Locator;
     public readonly accountCard: Locator;
 
-    constructor(public readonly page: Page, customer: components['schemas']['Customer'] ) {
-        this.customerId = customer.id;
-        this.customer = customer;
+    constructor(public readonly page: Page) {
         this.editButton = page.getByRole('button', { name: 'Edit' });
         this.generalTab = page.getByRole('link', { name: 'General' });
         this.accountCard = page.locator('.sw-customer-card')
     }
 
-    async goTo() {
-        await this.page.goto(`#/sw/customer/detail/${this.customerId}/base`);
+    async goTo({ customerId }: Record<string, unknown>) {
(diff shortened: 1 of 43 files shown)


src/page-objects/administration/FlowBuilderDetail.ts
```diff
@@ -0,0 +1,19 @@
...
+    public readonly generalTab: Locator;
+    public readonly flowTab: Locator;
+
+    constructor(public readonly page: Page) {
+        this.saveButton = page.locator('.sw-flow-detail__save');
+        this.generalTab = page.locator('.sw-flow-detail__tab-general');
+        this.flowTab = page.locator('.sw-flow-detail__tab-flow');
+    }
+
+    url(flowId: string, tabName? = 'general') {
```
pweyck:
```suggestion
    url(flowId: string, tabName: string = 'general') {
```

src/page-objects/administration/CustomerDetail.ts
```diff
@@ -1,24 +1,18 @@
...
-        this.customer = customer;
+    constructor(public readonly page: Page) {
         this.editButton = page.getByRole('button', { name: 'Edit' });
         this.generalTab = page.getByRole('link', { name: 'General' });
         this.accountCard = page.locator('.sw-customer-card')
     }
 
-    async goTo() {
-        await this.page.goto(`#/sw/customer/detail/${this.customerId}/base`);
+    url(customerId: string | undefined) {
```
pweyck:
```suggestion
    url(customerId: string) {
```

src/page-objects/administration/OrderDetail.ts
```diff
@@ -1,20 +1,18 @@
...
-        this.orderData = orderData;
+    constructor(public readonly page: Page) {
+        this.saveButton = page.locator('.sw-order-detail__smart-bar-save-button');
         this.dataGridContextButton = page.locator('.sw-data-grid__actions-menu').and(page.getByRole('button'));
         this.orderTag = page.locator('.sw-select-selection-list__item');
     }
 
-    async goTo() {
-        await this.page.goto(`#/sw/order/detail/${this.orderData.id}/general`);
+    url(orderId: string | undefined) {
```
pweyck:
```suggestion
    url(orderId: string) {
```

src/page-objects/administration/FlowBuilderDetail.ts
yusufttur: It would be nice if we could make the tab optional to visit any tab without a second click.

Phil23: If a variable has a default value, the type is inferred from that.

src/page-objects/administration/CustomerDetail.ts
Phil23: I added this because of an issue of our entity schema. The schema says that it can be undefined and if we want to pass the data from our API types to methods of the test suite we have to use this workaround. Or otherwise, fix the schema definition.

pweyck: Did you use the insert type or the type that you get on `GET`? On insert the id is optional

src/page-objects/administration/FlowBuilderDetail.ts
Phil23: @yusufttur The tab is already optional with my implementation. If you don't pass it, it will route to the general tab, which is the default.

src/page-objects/administration/CustomerDetail.ts
Phil23: It comes from the definitions of the API client that we use. This should be the line which adds the types.

```
const customer = (await customerResp.json()) as { data: components['schemas']['Customer'] };
```

pweyck: Could we not define a new type that extends the type and makes it `string`?

Phil23: Sure, but then we have to do this kind of overrides more and more and even expose them. That could end up in a rabbit hole of fixing the actual entity schema on our end.

yusufttur (dismissed):

src/page-objects/administration/CustomerDetail.ts
Phil23: @pweyck I added local overrides as discussed.

pweyck (dismissed):

pweyck (approved):

PR #37, opened 2024-06-20
First iteration of a new Test Data Service

Phil23:
The idea is to not use fixtures for test data creation, but a central service that is more flexible.

- No fixtures that depend on each other and cannot be edited from the ouside.
- More flexible and easy test data creation that can be adjusted to the test case.
- Automated clean up of created test data.

**Before:**

```TypeScript
test('My test scenario', async ({
    ShopCustomer,
    ProductData,
    StorefrontProductDetail,
}) => {
    await ShopCustomer.goesTo(StorefrontProductDetail.url(ProductData));
}
```

**After:**

```TypeScript
test('My test scenario', async ({
    ShopCustomer,
    TestDataService,
    StorefrontProductDetail,
}) => {
    const product = await TestDataService.createProductWithImage();
    
    await ShopCustomer.goesTo(StorefrontProductDetail.url(product));
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,14 +36,15 @@
     "prepublishOnly": "npm run build"
   },
   "dependencies": {
+    "@axe-core/playwright": "4.9.1",
     "@playwright/test": "^1.44.1",
     "@shopware/api-client": "0.5.0",
-    "@axe-core/playwright": "4.9.1",
     "axe-html-reporter": "2.2.3",
     "image-js": "0.35.5",
     "uuid": "9.0.1"
   },
   "devDependencies": {
+    "@types/node": "20.14.8",
     "@types/uuid": "9.0.8",
     "@typescript-eslint/eslint-plugin": "7.8.0",
     "@typescript-eslint/parser": "7.8.0",
--- a/src/services/TestDataService.ts
+++ b/src/services/TestDataService.ts
@@ -1,0 +1,1248 @@
+import { createRandomImage } from './ImageHelper';
+import type { AdminApiContext } from './AdminApiContext';
+import type { IdProvider } from './IdProvider';
+import type {
+    Product,
+    PropertyGroup,
+    Category,
+    Media,
+    Tag,
+    Rule,
+    Order,
+    Currency,
+    Customer,
+    CustomerAddress,
+    PaymentMethod,
+    ShippingMethod,
+    StateMachine,
+    StateMachineState,
+    Promotion,
+    SalesChannel,
+} from '../types/ShopwareTypes';
+
+export interface CreatedRecord {
+    resource: string;
+    id: string;
+}
+
+export interface SimpleLineItem {
+    product: Product;
+    quantity?: number;
+    position?: number;
+}
+
+export interface SyncApiOperation {
+    entity: string;
+    action: 'upsert' | 'delete';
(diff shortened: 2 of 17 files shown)


src/services/TestDataService.ts
```diff
@@ -0,0 +1,1234 @@
...
+     * @param shouldCleanUp - The config setting for the automated data clean up. Default is "true".
+     */
+    setCleanUp(shouldCleanUp = true) {
+        this.shouldCleanUp = shouldCleanUp;
+    }
+
+    /**
+     * Will delete all entities created by the data service via sync API.
+     */
+    async cleanUp() {
```
pweyck, line 807: Will be interesting to see, if this works for all entities. On principle it should work, if you delete all entities that were created in reverse order. I think some entities like orders, customers, sales channel might need custom logic.

tests/TestDataService.spec.ts
```diff
@@ -0,0 +1,26 @@
+import { test, expect } from '../src/index';
+
+test('Data Service', async ({
+    TestDataService,
+    DefaultSalesChannel,
+}) => {
+
+    const category = await TestDataService.createCategory(DefaultSalesChannel.salesChannel.navigationCategoryId, { name: 'Custom Category' });
```
pweyck: Could we use the `DefaultSalesChannel.salesChannel.navigationCategoryId` by default and maybe also move it into props? 
Something like `namePrefix` or `nameSuffix` would also be useful, the name could then be build by combining it with the unique id.

tests/TestDataService.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+    expect(category.name).toEqual('Custom Category');
+
+    const product = await TestDataService.createProductWithImage({ description: 'Test Description' });
+    expect(product.description).toEqual('Test Description');
+    expect(product.coverId).toBeDefined();
+
+    const customer = await TestDataService.createCustomer(DefaultSalesChannel.salesChannel, 'mr', { firstName: 'Luke', lastName: 'Skywalker' })
```
pweyck: Same thing with defaults and I'd also prefer props if all of them are optional

src/services/TestDataService.ts
Phil23: I had it as separate delete calls first and then changed it to one single sync API call. My hope was that within the sync call, it would resolve associations correctly. It seemed to work already for orders and customers.

tests/TestDataService.spec.ts
Phil23: I added more default settings to the constructor and made more parameters optional.

I also added a general default config for `namePrefix` and `nameSuffix` and added it to all entity names.

tests/TestDataService.spec.ts
Phil23: I added more defaults and put the overrides to the front. I'm not a big fan of parameters as objects if it is just a few parameters.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@shopware-ag/acceptance-test-suite",
-  "version": "2.0.1",
```
pweyck: Is there are reason you've increased it manually?

Phil23: No, I already changed it.

pweyck (approved):

PR #38, opened 2024-06-21
feat: make theme compilation compatible with SaaS instances

pweyck:
- fix waiting for theme compilation in SaaS
- add SetupSaaS task
- remove theme caching for now, because it does not work in all environments
- use `parallelIndex` instead of `workerIndex` to prevent unnecessary creation of different worker fixtures (for example when one test fails, it will use a different workerIndex after that)
- mock sbp calls, that fail for users that are created by the ats

### DIFF
--- a/src/fixtures/DefaultSalesChannel.ts
+++ b/src/fixtures/DefaultSalesChannel.ts
@@ -1,8 +1,7 @@
-import { test as base, expect, APIResponse } from '@playwright/test';
+import { test as base, expect } from '@playwright/test';
 import type { FixtureTypes } from '../types/FixtureTypes';
 import type { Customer, SalesChannel } from '../types/ShopwareTypes';
 import type { components } from '@shopware/api-client/admin-api-types';
-import { createHash } from 'crypto';
 import {
     getLanguageData,
     getCurrency,
@@ -13,6 +12,7 @@ import {
     getSnippetSetId,
     getThemeId,
 } from '../services/ShopwareDataHelpers';
+import { isSaaSInstance } from '../services/ShopInfo';
 
 interface StoreBaseConfig {
     storefrontTypeId: string;
@@ -36,7 +36,13 @@ export interface DefaultSalesChannelTypes {
         salesChannel: SalesChannel;
         customer: Customer;
         url: string;
-    }
+        themeSeed: string | null;
+    },
+    DefaultStorefront: {
+        salesChannel: components['schemas']['SalesChannel'];
+        customer: Customer;
+        url: string;
+    },
 }
 
 export const test = base.extend<NonNullable<unknown>, FixtureTypes>({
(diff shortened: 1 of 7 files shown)


src/fixtures/DefaultSalesChannel.ts
```diff
@@ -303,6 +301,51 @@ export const test = base.extend<NonNullable<unknown>, FixtureTypes>({
...
+                        break;
+                    }
+
+                    // eslint-disable-next-line playwright/no-wait-for-timeout
+                    await tmpPage.waitForTimeout(1000);
+                }
+            }
+
+            await use({
+                ...DefaultSalesChannel,
```
Phil23, line 348: Is there a reason why this is a separate fixture if it exposes the exact same data as the `DefaultSalesChannel`? Could be confusing.

src/fixtures/PageContexts.ts
```diff
@@ -76,19 +79,59 @@ export const test = base.extend<FixtureTypes>({
...
 
         await use(page);
 
         await page.close();
         await context.close();
     },
 });
+
+
+async function mockApiCalls(page: Page) {
```
Phil23: We should extract that to a separate file.

src/fixtures/PageContexts.ts
```diff
@@ -76,19 +79,59 @@ export const test = base.extend<FixtureTypes>({
...
...monthlyPlanSurchargeAmount":0,"salesChannel":"sbp_frontend","bookedCommercialEdition":{"id":3,"name":"beyond","label":"beyond","tags":[{"id":3,"name":"REGULAR"}],"basePrice":6500,"includedGmv":15000000,"includedOrderAmount":15000,"sortPriority":3,"isRegular":true,"isUpgrade":false,"isMigration":false,"isClient":false,"isReseller":false,"displayName":"Shopware Beyond"},"details":{"id":10342,"mode":"gmv","name":"beyond","label":"beyond","tags":[{"id":3,"name":"REGULAR"}],"basePrice":6500,"chargingInterval":1,"monthlySurchargeIgnored":false,"includedAmount":15000000,"plannedAmount":2000000,"discountAmount":6825,"discountExpirationDate":"2025-06-17T00:00:00+02:00","pricingPeriods":[{"id":10819,"includedAmount":15000000,"plannedAmount":2000000,"discountAmount":81900,"relativeDiscountAmount":100,"periodContractValue":0,"plannedAmountFixed":true,"planBasePrice":81900,"planPrice":0,"planMonthlyPrice":0,"periodTotalPrice":0,"periodMonthlyPrice":0,"bookingPrice":0,"startDate":"2024-06-18 00:00:00","endDate":"2025-06-17 23:59:59","paasDefinition":null,"consultingServiceDefinition":null,"paasPrice":0,"consultingServicePrice":0}],"calculatedPrice":0},"chargingInterval":1,"calculatedPrice":0}}',
```
Phil23: I'm a bit concerned about exposing internal software APIs in a public package. For example, is the `companyId` the real id from the account in the SBP?

This shouldn't be part of the NPM package that can be used as public dev tooling. What do you think?

src/tasks/saas-tasks.ts
```diff
@@ -0,0 +1,38 @@
+import test, { expect } from '@playwright/test';
+import { isSaaSInstance } from '../services/ShopInfo';
+import { FixtureTypes } from '../types/FixtureTypes';
+import { Task } from '../types/Task';
+
+export const SetupInstance = test.extend<{ SetupInstance: Task } & FixtureTypes>({
```
Phil23: When is this task supposed to be used? Does it always have to be called if a SaaS instance is the target?

src/fixtures/DefaultSalesChannel.ts
pweyck: It's just the default sales channel + theme compilation. So if you just request the sales channel no theme compilation needs to happen

src/fixtures/PageContexts.ts
pweyck: I'll try to remove some stuff like ids, because they should not matter to the admin

src/tasks/saas-tasks.ts
pweyck: Yes, because saas instances need to be setup, before they can be used. You can see that in the platform MR

Phil23: Should we make it a simpler fixture then and not a task. Because it is not supposed to be a task of an actor but a simple system call. Then you also don't have to use the ugly double function call.

pweyck: Alright, I'll refactor it into a fixture

src/services/ShopwareDataHelpers.ts
```diff
@@ -1,242 +1,242 @@
...
+            }],
+        },
+    });
+
+    const result = (await resp.json()) as { data: { id: string }[]; total: number };
+
+    return result.data[0].id;
+};
+
+export const getCurrency = async (isoCode: string, adminApiContext: AdminApiContext) => {
```
Phil23, line 46: What is the reason to remove the default here? Technically a break. 😅

Phil23 (approved):

PR #70, opened 2024-07-26
NEXT-37447 - Create manufacturer with image test data

frobel:
As an ATS user, I want to create a Manufacturer with an image via the TestDataService so that I can use it to test my KPIs within an analytics app.


Phil23:
Please add the test for the creation. The rest is fine. 👍

Phil23 (approved):

PR #78, opened 2024-08-07
feat: Add possibility to use custom customer for Login task

frobel:


src/services/TestDataService.ts
```diff
@@ -473,7 +473,10 @@ export class TestDataService {
             data: basicCustomerStruct,
         });
 
-        const { data: customer } = (await response.json()) as { data: Customer };
+        const customerData = (await response.json()) as { data: Customer };
+
+        // @ts-ignore
+        const customer: Customer = { ...customerData.data, password: basicCustomerStruct.password };
```
frobel: Someone now how to avoid the message which pops up from `password: basicCustomerStruct.password`

```
TS2322: Type string | undefined is not assignable to type string
Type undefined is not assignable to type string
```

src/tasks/shop-customer/Account/Login.ts
```diff
@@ -1,12 +1,19 @@
...
     Login: async ({ ShopCustomer, DefaultSalesChannel, StorefrontAccountLogin, StorefrontAccount }, use)=> {
-        const task = () => {
+        const task = (customCustomer?: Customer) => {
             return async function Login() {
-                const { customer } = DefaultSalesChannel;
+
+                let customer: Customer;
+                if (!customCustomer) {
+                    customer = DefaultSalesChannel.customer;
+                } else {
```
yusufttur, line 14:
```suggestion
               customer = customCustomer ? customCustomer : DefaultSalesChannel.customer;
```

src/tasks/shop-customer/Account/Login.ts
```diff
@@ -1,12 +1,19 @@
...
+                    customer = customCustomer;
+                }
```
yusufttur:
```
             customer = customCustomer ? customCustomer : DefaultSalesChannel.customer;
```

src/services/TestDataService.ts
SebastianFranze: You have to ensure that the password exists... Because it can be `null`/`undefined` for guest customers. A simple check before this line should do the trick.

```ts
if (typeof basicCustomerDataStruc.password !== 'string')  {
    // fail the test
}
```

src/tasks/shop-customer/Account/Login.ts
```diff
@@ -1,12 +1,14 @@
 import { test as base } from '@playwright/test';
 import type { Task } from '../../../types/Task';
 import type { FixtureTypes} from '../../../types/FixtureTypes';
+import { Customer } from '../../../types/ShopwareTypes';
```
SebastianFranze:
```suggestion
import type { Customer } from '../../../types/ShopwareTypes';
```

src/tasks/shop-customer/Account/Login.ts
```diff
@@ -1,12 +1,14 @@
...
 
 export const Login = base.extend<{ Login: Task }, FixtureTypes>({
     Login: async ({ ShopCustomer, DefaultSalesChannel, StorefrontAccountLogin, StorefrontAccount }, use)=> {
```
SebastianFranze:
```suggestion
    Login: async ({ ShopCustomer, DefaultSalesChannel, StorefrontAccountLogin, StorefrontAccount }, use) => {
```

src/services/TestDataService.ts
SebastianFranze: Or even easier... You cast it to a string:

```suggestion
        const customer: Customer = { ...customerData.data, password: String(basicCustomerStruct.password) };
```

But this also means if there is no password set you try to login with the string "undefined" :sweat_smile:

SebastianFranze (approved):

PR #79, opened 2024-08-08
feat: Improve MailpitApiContext 

yusufttur:
This MR contains some improvements for MailpitApiContext:
- adds email parameter to the `getEmailBody()`,  `getRenderMessageTxt()` and `getLinkFromMail()` methods to return intended mail content in each call in parallel testing.
- improves descriptive comments for the methods.

### DIFF
--- a/src/services/MailpitApiContext.ts
+++ b/src/services/MailpitApiContext.ts
@@ -12,58 +12,60 @@ export interface Email {
 }
 
 export class MailpitApiContext {
-    public context: APIRequestContext;
-  
-    constructor(context: APIRequestContext) {
-        this.context = context;
-    }
-  
-    /**
-     * Fetches email headers based on the recipient's email address.
-     * @param email - The email address of the recipient.
-     * @returns An Email object containing the email headers.
-     */
-    async getEmailHeaders(email: string): Promise<Email> {
-      const response = await this.context.get('/api/v1/search', {
-        params: {
-          kind: 'To.Address',
-          query: email,
-        },
-      });
-  
-      const responseJson = await response.json();
-    
-      return {
-        fromName: responseJson.messages[0].From.Name,
-        fromAddress: responseJson.messages[0].From.Address,
-        toName: responseJson.messages[0].To[0].Name,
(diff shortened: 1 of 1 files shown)


frobel:
I'm not 100% sure, but is that not a breaking change if we add a parameter (mandatory) to an already existing method?

src/services/MailpitApiContext.ts
```diff
@@ -12,118 +12,122 @@ export interface Email {
...
+      emailId: message.ID,
+    };
+  }
+
+  /**
+   * Retrieves the body content of the email as an HTML string.
+   * @param email - The email address of the recipient.
+   * @returns A promise that resolves to the HTML content of the email.
+   */
+  async getEmailBody(email: string): Promise<string> {
```
frobel, line 52: Is that a breaking change?

frobel (changes requested):

src/services/MailpitApiContext.ts
yusufttur: Actually, not, because we haven't used these methods before. 😄

frobel: we not, but maybe external you don't know yet ;)

yusufttur: Idk if someone uses this method. But there was a problem if we didn't add an email parameter. Because `getEmailBody()` brings the latest email content, we need to use it with `generateEmailContent(email: string)` to get the email body and its matching email headers. If we would like to use it alone to get the latest email body, like in the first version of the method, then we need to use `latest` as the email parameter. 
Docs: https://mailpit.axllent.org/docs/api-v1/view.html#get-/view/-ID-.html

frobel: yes that is okay...we should then define within the commit message that this is a breaking change.

https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-description-and-breaking-change-footer

yusufttur: I added a commit message with description and breaking change footer

frobel (approved):

PR #85, opened 2024-08-13
feat: admin - Add notification await

Deristes:
In some new tests we want to await for a notification to appear.
Let's create a general function to handle this case.

### DIFF
--- a/src/tasks/shop-admin/ExpectNotification.ts
+++ b/src/tasks/shop-admin/ExpectNotification.ts
@@ -1,0 +1,21 @@
+import { test as base } from '@playwright/test';
+import type { Task } from '../../types/Task';
+import type { FixtureTypes } from '../../types/FixtureTypes';
+
+export const ExpectNotification = base.extend<{ ExpectNotification: Task }, FixtureTypes>({
+    ExpectNotification: async ({ ShopAdmin }, use ) => {
+        const task = (message: string, close = true) => {
+            return async function ExpectNotification() {
+                const notification = ShopAdmin.page.locator('.sw-alert', { hasText: message });
+                await ShopAdmin.expects(notification).toBeVisible();
+                await ShopAdmin.expects(notification).toContainText(message);
+
+                if (close) {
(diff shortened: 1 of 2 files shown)


src/tasks/shop-admin/ExpectNotification.ts
```diff
@@ -0,0 +1,20 @@
+import { test as base } from '@playwright/test';
+import type { Task } from '../../types/Task';
+import type { FixtureTypes} from '../../types/FixtureTypes';
```
Phil23:
```suggestion
import type { FixtureTypes } from '../../types/FixtureTypes';
```

src/tasks/shop-admin/ExpectNotification.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const ExpectNotification = base.extend<{ ExpectNotification: Task }, FixtureTypes>({
+    ExpectNotification: async ({ ShopAdmin }, use ) => {
+        const task = (message: string, options: { close: boolean } = { close: true }) => {
```
Phil23: As there is only one option, I would prefer exposing the close option as a direct function parameter.

src/tasks/shop-admin/ExpectNotification.ts
```diff
@@ -0,0 +1,20 @@
...
+            return async function ExpectNotification() {
+                const notification = ShopAdmin.page.locator('.sw-alert', { hasText: message });
+                await ShopAdmin.expects(notification).toBeVisible();
+                await ShopAdmin.expects(notification).toContainText(message);
+                if (options.close) {
+                    await notification.locator('.sw-alert__close').click();
+                }
```
Phil23:
```suggestion
                await ShopAdmin.expects(notification).toContainText(message);

                if (options.close) {
                    await notification.locator('.sw-alert__close').click();
                }
```

Phil23 (changes requested):

Phil23 (approved):

PR #98, opened 2024-08-29
NEXT-33825 - Add the postal code on registration

nguyenytran:
In this form https://github.com/shopware/shopware/issues/4222, an updated postal code value is required when registering.


yusufttur:
You need to use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
Commit should start with "feat:" in your case, as it is a minor change.

PR #99, opened 2024-08-29
feat: Add the postal code on registration

nguyenytran:
In this ticket https://github.com/shopware/shopware/issues/4222, an updated postal code value is required when registering.

### DIFF
--- a/src/page-objects/storefront/AccountLogin.ts
+++ b/src/page-objects/storefront/AccountLogin.ts
@@ -18,6 +18,7 @@ export class AccountLogin implements PageObject {
     public readonly streetAddressInput: Locator;
     public readonly cityInput: Locator;
     public readonly countryInput: Locator;
+    public readonly postalCodeInput: Locator;
     public readonly registerButton: Locator;
 
     constructor(public readonly page: Page) {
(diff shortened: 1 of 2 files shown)


src/page-objects/storefront/AccountLogin.ts
```diff
@@ -35,6 +36,7 @@ export class AccountLogin implements PageObject {
         this.streetAddressInput = this.billingAddressFormArea.getByLabel('Street address*');
         this.cityInput = this.billingAddressFormArea.getByLabel('City*');
         this.countryInput = this.billingAddressFormArea.getByLabel('Country*');
+        this.postalCodeInput = this.billingAddressFormArea.getByLabel('Postal code');
```
yusufttur: I guess now postal code is a mandatory area, but why don't we add "*" to this snippet?

nguyenytran: It depends on the country to be mandatory or not, since not sure I just getByLabel('Postal code') instead of getByLabel('Postal code*')

yusufttur: But if it is mandatory in a specific country, does it get an "*" at the end?

nguyenytran: @yusufttur but i have not tested it to see if it works correctly and i do not know how to test it

nguyenytran: But if it is mandatory in a specific country, does it get an "*" at the end? yes, do i need to update it

yusufttur: Yes, please update and add "*" since it is a default mandatory.

yusufttur (approved):

PR #105, opened 2024-09-02
fix: add last price range with infinity validity

frobel:


GitEvil (dismissed):

src/services/TestDataService.ts
```diff
@@ -1192,6 +1192,11 @@ export class TestDataService {
                 price: p(80, 67.23),
                 quantityStart: 21,
                 quantityEnd: 50,
+            }, {
+                ruleId: ruleId,
+                price: p(70, 58.82),
+                quantityStart: 51,
+                quantityEnd: null,
```
yusufttur, line 1199:
```suggestion
                quantityEnd:,
```
Leave blank to set the upper limit to infinity.

frobel: it is not possible with the current definition of Price Array

GitEvil (approved):

PR #110, opened 2024-09-05
feat: add possibility to create variants out of property groups

frobel:


tests/TestDataService.spec.ts
```diff
@@ -95,6 +95,17 @@ test('Data Service', async ({
...
     expect(rule.description).toEqual('Rule description');
 
+    const parentProduct = await TestDataService.createBasicProduct();
+    const propertyGroups: PropertyGroup[] = [];
+    const propertyGroupColor = await TestDataService.createColorPropertyGroup();
+    const propertyGroupText = await TestDataService.createTextPropertyGroup();
+    propertyGroups.push(propertyGroupColor);
+    propertyGroups.push(propertyGroupText);
+
+    const variantProducts = await TestDataService.createVariantProducts(parentProduct, propertyGroups, { description: 'Variant description'});
```
yusufttur, line 105: Why do you need to add a variant description? 🤔

frobel: I check with that, if the override function works so that you can define some specific variant values

yusufttur (approved):

PR #118, opened 2024-09-11
fix: make ATS compatible with 6.5.x

pweyck:


tests/TestDataService/OrderCreation.spec.ts
```diff
@@ -0,0 +1,43 @@
+import { test, expect, type Order, APIResponse } from '../../src/index';
+
+
+test('Order creation with TestDataService', async ({
```
pweyck, line 4: I've moved this into it's own test, because the spec file got way to big. We should not write tests that are bigger than 100 lines of code.

src/services/TestDataService.ts
```diff
@@ -144,6 +145,7 @@ export class TestDataService {
         const productResponse = await this.AdminApiClient.post('./product?_response=detail', {
             data: basicProduct,
         });
+        expect(productResponse.ok()).toBeTruthy();
```
pweyck, line 148: We should always check the response, if we read the data from the response. Otherwise we get a low level runtime error, which is harder to debug.

src/page-objects/storefront/AccountAddresses.ts
```diff
@@ -9,7 +9,7 @@ export class AccountAddresses implements PageObject {
     public readonly useDefaultShippingAddressButton: Locator;
 
     constructor(public readonly page: Page) {
-        this.addNewAddressButton = page.getByRole('link', { name: 'Add new address' });
+        this.addNewAddressButton = page.getByRole('link', { name: /Add (new )?address/ });
```
frobel, line 12: What happend here?

src/page-objects/storefront/Category.ts
```diff
@@ -13,6 +13,6 @@ export class Category implements PageObject {
     }
 
     url(categoryName: string): string {
-        return `${categoryName}`;
+        return `${categoryName}/`;
```
frobel, line 16: and here? Why is there a `/`

src/services/TestDataService.ts
frobel: @Phil23 do you know why you didn't add there expect here?

src/services/TestDataService.ts
```diff
@@ -1749,7 +1781,6 @@ export class TestDataService {
         const totalPrice = unitPrice * (lineItem.quantity || 1);
 
         const basicPromotionLineItemStruct = {
-            promotionId: promotion.id,
```
frobel: Why do you delete that `promotionId` field?

src/services/TestDataService.ts
```diff
@@ -702,16 +717,18 @@ export class TestDataService {
     async createBasicShippingMethod(
         overrides: Partial<ShippingMethod> = {},
     ): Promise<ShippingMethod> {
+        const deliveryTime = await this.getAllDeliveryTimeResources()
 
-        const deliveryTime = await this.getAllDeliveryTimeResources();
-
+        overrides.availabilityRuleId ??= (await this.getRule('Always valid (Default)')).id
```
frobel, line 722: Is that 6.5.x relevant? I see in the docs its is [required in 6.5](https://shopware.stoplight.io/docs/admin-api/branches/v6.5/32dcb684349f4-create-a-new-shipping-method-resources). and optional in 6.6.

frobel (dismissed):

src/services/TestDataService.ts
pweyck: yes, because it's required in 6.5, we should provide the default

src/services/TestDataService.ts
pweyck: it's not allowed to provide the id in 6.5 for some reason

src/page-objects/storefront/Category.ts
pweyck: weird bug in 6.5, without the `/` it does not work, should work both ways in 6.6

src/page-objects/storefront/AccountAddresses.ts
pweyck: different snippet in 6.5 `Add new address`, so we match both

src/services/TestDataService.ts
pweyck: I'll double check this, because I'm not entirely sure how it works

frobel: @pweyck I had trouble to create a Promotion via AdminAPI and ask Soner would could be the problem. The solution was: https://gitlab.shopware.com/shopware/6/product/platform/-/commit/bc1be7402fff6104adf71194c056d240b640192d

frobel: The solution is to backport the MR from 6.6. to 6.5.

pweyck: I've backported the fix and will revert this change

PR #124, opened 2024-09-13
set type: module

amenk:
Without this change I got

```
// BaseTestFile.ts
import { test as base } from '@shopware-ag/acceptance-test-suite';
                                                                  
SyntaxError: Cannot use import statement outside a module
```

I am not sure if that is the proper solution, so it's open for discussion


shyim:
We could use here https://github.com/isaacs/tshy , so the result package is common js and module js 🤔

amenk:
Not sure if I understand.

I had to add this to the package.json of *my* project which uses the test-suite plugin and was created as described in the README

The `BaseTestFile.ts` is located in my project.
I dont build any dist of my test-project.

frobel (approved):

PR #126, opened 2024-09-13
Change AccountOrder orderExpandButton to locator

tobiasberge:
image

For accessibility improvements we want to change the translation from generic "Expand" wording to "Show details" or "Hide details".

To make it still work in older versions that don't have the updated snippet, use a class selector instead of text.

### DIFF
--- a/src/page-objects/storefront/AccountOrder.ts
+++ b/src/page-objects/storefront/AccountOrder.ts
@@ -8,7 +8,7 @@ export class AccountOrder implements PageObject {
     public readonly digitalProductDownloadButton: Locator;
 
     constructor(public readonly page: Page) {
-        this.orderExpandButton = page.locator('.order-hide-btn').first();
+        this.orderExpandButton = page.getByRole('button', {name: /Expand|Show details/}).first();
         this.cartLineItemImages = page.locator('.line-item-img-link');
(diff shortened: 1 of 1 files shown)


src/page-objects/storefront/AccountOrder.ts
```diff
@@ -8,7 +8,7 @@ export class AccountOrder implements PageObject {
     public readonly digitalProductDownloadButton: Locator;
 
     constructor(public readonly page: Page) {
-        this.orderExpandButton = page.getByRole('button', {name: 'Expand'}).first();
+        this.orderExpandButton = page.locator('.order-hide-btn').first();
```
pweyck: Usually, we want to prefer aria selectors (+ text) instead of classes. You can use a regex to match the old and new string. 
```suggestion
        this.orderExpandButton = page.getByRole('button', {name: /Expand|Show details/}).first();
```

pweyck, line 11:
```suggestion
        this.orderExpandButton = page.getByRole('button', {name: /Expand|Show details/}).first();
```

pweyck (approved):

src/page-objects/storefront/AccountOrder.ts
tobiasberge: Ah nice trick, didn't know that this is possible here. 👍

PR #127, opened 2024-09-16
tests: skip tests using mailpit when running against a SaaS instance

philipreinken:
Some tests are checking for E-Mails using a local mailpit instance; this is not possible when running against a remote and/or SaaS instance.

### DIFF
--- a/tests/Mailpit.spec.ts
+++ b/tests/Mailpit.spec.ts
@@ -7,7 +7,11 @@ test('Verify sign-up email', { tag: '@Email' }, async ({
     page,
     MailpitApiContext,
     TestDataService,
+    InstanceMeta,
 }) => {
+    // eslint-disable-next-line playwright/no-skipped-test
+    test.skip(InstanceMeta.isSaaS, 'Skipping test because it requires a local mailpit instance.');
+
     // Login with a new customer
     const email = IdProvider.getIdPair().id + '@test.com';
     const password = 'shopware';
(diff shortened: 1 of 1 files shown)


philipreinken:
Depends on #126

mkraeml (dismissed):

src/page-objects/storefront/AccountOrder.ts
```diff
@@ -8,7 +8,7 @@ export class AccountOrder implements PageObject {
     public readonly digitalProductDownloadButton: Locator;
 
     constructor(public readonly page: Page) {
-        this.orderExpandButton = page.getByRole('button', {name: 'Expand'}).first();
+        this.orderExpandButton = page.locator('.order-hide-btn').first();
```
pweyck: Is this buggy? Usually, we want to prefer aria selectors + text instead of classes, like in this example.

philipreinken: This is from @tobiasberge 's PR here: #126 

I rebased on top of that, because the tests were failing without it.

pweyck: https://github.com/shopware/acceptance-test-suite/pull/126/files#r1766818473

tobiasberge: This was failing because the translation was changed from "Expand" to "Show details". And when just changing the name I guess it would stop working in older SW versions that don't have the updated translation.

pweyck: We can merge the other PR first and merge this after a rebase

pweyck (approved):

PR #129, opened 2024-09-26
feat: redefine page and context from AdminPage

pweyck:
The `page` and `context` fixtures are provided by playwright.

Right now they are fresh objects that are unrelated to the ATS fixture `AdminPage`. This causes some issue if you use the page or context object in your tests because you might expect them to be the same. For example when waiting for page events.

To solve this we redefine them as being the `AdminPage` and `AdminPage.context()` respectively.


seggewiss (approved):
LGTM!

PR #156, opened 2024-11-08
NEXT-33515 - Rule preview test

GitEvil:
adds fixtures for Rule Preview test

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@shopware-ag/acceptance-test-suite",
-  "version": "3.11.0",
+  "version": "4.0.0",
   "description": "Shopware Acceptance Test Suite",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@shopware-ag/acceptance-test-suite",
-  "version": "3.11.0",
+  "version": "3.12.1",
```
vanpham-sw: Version should be either 3.11.4 or 4.0.0. I'm not sure if we need your changes in both versions, how would we do it, cherry pick with another PR?

sguder (dismissed):

sguder (approved):

ravenokko (approved):

PR #174, opened 2024-11-15
feat: add custom field page objects

frobel:


src/page-objects/administration/CustomField.ts
```diff
@@ -0,0 +1,37 @@
+import type { Page, Locator } from '@playwright/test';
+import type { PageObject } from '../../types/PageObject';
+
+export class CustomField implements PageObject {
+    public readonly addNewSetButton: Locator;
+    public readonly customFieldRows: Locator;
+
+    constructor(public readonly page: Page) {
+        this.addNewSetButton = page.getByRole('button', { name: 'Add new set' });
+        this.customFieldRows = page.getByRole('row');
```
kstala: @frobel Is this the only way to use a locator like this one?

kstala (dismissed):

src/page-objects/administration/CustomField.ts
frobel: I want to provide with that solution the community to get all rows via a property so that you can count or dig deeper with that locator.

frobel: to actually work with the row elements it is better to use the method: `getLineItemByCustomFieldName()` (see below)

vanpham-sw (dismissed):

src/page-objects/administration/CustomFieldDetail.ts
```diff
@@ -0,0 +1,77 @@
...
+    public readonly customFieldLabelEnglishGBInput: Locator;
+    public readonly customFieldPlaceholderEnglishGBInput: Locator;
+    public readonly customFieldHelpTextEnglishGBInput: Locator;
+    public readonly customFieldDeleteListButton: Locator;
+    public readonly customFieldDeleteDialog: Locator;
+    public readonly customFieldDeleteCancelButton: Locator;
+    public readonly customFieldDeleteButton: Locator;
+
+    constructor(public readonly page: Page) {
+        super(page);
```
vanpham-sw, line 24: just curious, why do you use **super(page);** in some pages

vanpham-sw (dismissed):

src/page-objects/administration/CustomFieldListing.ts
```diff
@@ -0,0 +1,37 @@
...
+    public readonly customFieldRows: Locator;
+
+    constructor(public readonly page: Page) {
+        this.addNewSetButton = page.getByText('Add new set');
+        this.customFieldRows = page.getByRole('row');
+    }
+
+    async getLineItemByCustomFieldSetName(customFieldName: string): Promise<Record<string, Locator>> {
+        const lineItem = this.page.getByRole('row').filter({ hasText: customFieldName });
+        const customFieldSetNameText = lineItem.locator('.sw-data-grid__cell--warehouse-name');
```
yusufttur: Could this locator be wrong? It says warehouse-name. 🤔

yusufttur (dismissed):

src/page-objects/administration/CustomFieldDetail.ts
frobel: I extend from the  `CustomFieldCreate` page-objects and with `super()` I tell the root objects the extended properties.

src/page-objects/administration/CustomFieldListing.ts
frobel: jep...i changed it

vanpham-sw (approved):

PR #179, opened 2024-11-21
fix: add openLandingPageByName

vanpham-sw:


src/page-objects/administration/Categories.ts
```diff
@@ -85,6 +85,11 @@ export class Categories implements PageObject {
     this.openInNewTabCheckbox = page.getByRole('checkbox', { name: 'Open in new tab' });
   }
 
+  async openLandingPageByName(landingPageName: string) {
```
frobel: Honestly I wouldn't add logic to the page-object if it is not necessary. For that we have the task objects. Here we wanna just add Locator logic. Therefore please use the task for `openLandingPageByName()`

frobel (changes requested):

src/page-objects/administration/Categories.ts
vanpham-sw: maybe a task in this case is overkill? I changed to:
`async getLandingPageByName(landingPageName: string) : Promise<Locator> {
    return this.landingPageItems.locator(`text="${landingPageName}"`);
  }`

frobel (approved):

PR #192, opened 2024-11-27
feat: Pages and tasks for product bulk edit

sguder:


src/tasks/shop-admin/Product/BulkEditProducts.ts
```diff
@@ -0,0 +1,123 @@
...
+                        const responsePromise = AdminProductBulkEdit.page.waitForResponse(response =>
+                            response.url().includes(`search/sales-channel`) && response.status() === 200 && response.request().method() === 'POST'
+                        );
+                        await AdminProductBulkEdit.salesChannelInput.fill(changes['saleschannel'].value);
+
+                        // Wait for the search call to be completed
+                        const response = await responsePromise;
+                        expect(response.ok()).toBeTruthy();
+                        const searchResult = await AdminProductBulkEdit.getDropdownEntry(changes['saleschannel'].value);
+                        await searchResult.click();
```
frobel, line 110: I don't know if that leads to flakiness because you are dependent on the search. And maybe the test will be slow. I'm not sure.

Another solution could be:

// AdminCustomFieldCreate Page Object

```
this.assignToSelectionList = page.locator('.sw-field')
            .filter({ hasText: 'Assign To' })
            .locator('.sw-select__selection');
        this.resultAssignToPopoverItemList = page.locator('.sw-select-result-list__content').getByRole('listitem');
```

// Platform Test - CreateCustomFieldSet
```
await AdminCustomFieldCreate.assignToSelectionList.click();
await AdminCustomFieldCreate.resultAssignToPopoverItemList.getByText('Products').click();
```

src/tasks/shop-admin/Product/BulkEditProducts.ts
```diff
@@ -0,0 +1,123 @@
...
+                    } 
+                }
+
+                await AdminProductBulkEdit.applyChangesButton.click();
+                await AdminProductBulkEdit.confirmModalApplyChangesButton.click();
+                await expect(AdminProductBulkEdit.confirmModalSuccessHeader).toBeVisible();
```
frobel: Between these two commands, there will be an operation. Should we check properly if the operation is done? Maybe the current implementation could lead to flakiness.

sguder: Added a check for the loading spinner to appear and disappear again. I think we can't wait for api calls, because we don't know at this point, which calls are fired. There are multiple depending on which fields are changed

src/tasks/shop-admin/Product/BulkEditProducts.ts
sguder: Doesn't this only pick an entry from the dropdown list? What if the entry, we want, isn't on that list? It is limited to show only a part of all entries

frobel: `await AdminCustomFieldCreate.resultAssignToPopoverItemList.getByText('Products').click();`

with that you can decide in your test, what entity you want to pick. In my example I wanted the "Products" entity.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@shopware-ag/acceptance-test-suite",
-  "version": "5.0.0",
+  "version": "5.2.0",
```
frobel: can be removed from your PR. The repo manage it by themself

src/types/ShopwareTypes.ts
```diff
@@ -71,7 +73,7 @@ export type Media = components['schemas']['Media'] & {
 }
 
 export type Tag = components['schemas']['Tag'] & {
-    id: string,
+    id: string;
```
frobel: please add a comma here

sguder: fixed

frobel (approved):

PR #193, opened 2024-11-28
feat: customer bulk edit

vanpham-sw:


src/tasks/shop-admin/Customers/BulkEditCustomers.ts
```diff
@@ -0,0 +1,76 @@
...
+                    await AdminCustomerBulkEdit.changeTypeSelect.click();
+                    await AdminCustomerBulkEdit.page.getByText(tagData.changeType).click();
+                    for (const tag of tagData.tags) {
+                        await AdminCustomerBulkEdit.enterTagsSelect.click();
+                        await AdminCustomerBulkEdit.page.getByText(tag).click();
+                    }
+                }
+                if (customFieldData) {
+                    const customFieldSet = await AdminCustomerBulkEdit.getCustomFieldLinkByName(customFieldData.customFieldSetName);
+                    while (!await customFieldSet.isVisible()) {
```
frobel, line 60: Why you chose that approche and not directly find the customSetName by Name and click it?

src/tasks/shop-admin/Customers/BulkEditCustomers.ts
```diff
@@ -0,0 +1,76 @@
...
+                        await AdminCustomerBulkEdit.customFieldArrowRightButton.click();
+                    }
+                    await customFieldSet.click();
+                    await AdminCustomerBulkEdit.customFieldCheckbox.click();
+                    await AdminCustomerBulkEdit.customFieldInput.fill(customFieldData.customFieldValue);
+                }
+                await AdminCustomerBulkEdit.applyChangesButton.click();
+                await AdminCustomerBulkEdit.confirmModalApplyChangesButton.click();
+                await expect(AdminCustomerBulkEdit.confirmModalSuccessHeader).toBeVisible();
```
frobel: can you here also use `ShopAdmin.expects` ...

src/services/TestDataService.ts
```diff
@@ -1111,6 +1112,22 @@ export class TestDataService {
         return salesChannel;
     }
 
+    async addSalesChannelLanguage(salesChannelId: string, languageCode: string) {
```
ravenokko: I would change to `assignSalesChannelLanguage` to match the ones above

src/services/TestDataService.ts
```diff
@@ -1111,6 +1112,22 @@ export class TestDataService {
...
+        const language = await getLanguageData(languageCode, this.AdminApiClient);
```
ravenokko: Should we really call the `getLanguageData` function here? I get that it's nice to only have to know the isoCode, but in my scenario at least, it would mean I'd have to call this function more than once because I need the languageId for other things, too.

ravenokko (changes requested):

src/services/TestDataService.ts
vanpham-sw: I'm using add trying to describe that I'm adding another language to the sales channel, which already has a default language.

src/services/TestDataService.ts
vanpham-sw: I don't get your expectation, maybe share your scenario here would help?

Let me explain why using the languageCode here: What I need is the language.id, but it can be dynamic because it's an id. On the other hand, the isoCode does not change that frequently, so I can base on the code to get whatever id it will be at the moment the test is executing.

src/tasks/shop-admin/Customers/BulkEditCustomers.ts
vanpham-sw: There can be some leftover customsets from other tests that's hiding mine, so it's kinda a precautionary step.

frobel: mhh I my case I could manage that directly with the following function:

  ```
await ShopAdmin.expects(customFieldSetTabContent.customFieldSetTab).toBeVisible();
  await ShopAdmin.expects(customFieldSetTabContent.customFieldSetTab).toHaveText(customFieldSetName);
  await customFieldSetTabContent.customFieldSetTab.click();
```

But you solution seems to be OK too...so up to you. You can resolve the thread if you like.

File: src/services/TestDataService.ts (hunk 3 above)
Comment (frobel) [reply src/services/TestDataService.ts]: I would also recommend to stick to our language slang with `assign` instead of `add`

File: src/tasks/shop-admin-tasks.ts [hunk 5]
```diff
@@ -3,9 +3,11 @@ import { mergeTests } from '@playwright/test';
 import { SaveProduct } from './shop-admin/Product/SaveProduct';
 import { ExpectNotification } from './shop-admin/ExpectNotification';
 import { CreateLinkTypeCategory } from './shop-admin/Category/CreateLinkTypeCategory';
+import { BulkEditCustomers } from './shop-admin/Customers/BulkEditCustomers';
 
 export const test = mergeTests(
     SaveProduct,
     ExpectNotification,
     CreateLinkTypeCategory,
+    BulkEditCustomers
```
frobel: comma missing

src/page-objects/administration/CustomerBulkEdit.ts
```diff
@@ -0,0 +1,91 @@
...
+  async getCustomFieldLinkByName(customFieldSetName: string): Promise<Locator> {
+    return this.page.locator('a').filter({
+      hasText: `${customFieldSetName}`,
+    });
+  }
+
+  url(): string {
+    return '#/sw/bulk/edit/customer';
+  }
+
```
frobel, line 89: empty line

src/page-objects/administration/CustomerBulkEdit.ts
```diff
@@ -0,0 +1,91 @@
+import { PageObject } from '../../types/PageObject';
+import { Locator, Page } from '@playwright/test';
+
+
```
frobel: emptyLine

frobel (changes requested):

src/services/TestDataService.ts
ravenokko: So in my scenario I need to:

1) Assign a language to a sales channel
2) Create a Sales Channel domain with that new language

For both steps I need the `languageId`. So for me it makes the most sense to write something like:

```
const language = getLanguageData(languageCode);
const languageId = language.id;

await TestDataService.assignSalesChannelLanguage(salesChannelId, languageId);
await TestDataService.createSalesChannelDomain(salesChannelId, languageId, x, y, z);
```

So in my case, I'd end up calling `getLanguageData()` at least twice because currently, assignSalesChannelLanguage calls it again because it asks for the language isoCode and not the languageId. For a similar case, if you look at `assignSalesChannelCurrency()` you'll see that `getCurrency()` is a separate call.

So what I'd want is:
1. Change `languageCode` to `languageId` in the `assignSalesChannelLanguage()` function arguments
2. remove `getLanguageData()` function to a separate call

ravenokko (dismissed):

src/services/TestDataService.ts
```diff
@@ -1185,6 +1234,26 @@ export class TestDataService {
         return salesChannel;
     }
 
+    /**
+     * Assigns a language to a sales channel.
+     *
+     * @param salesChannelId - The uuid of the sales channel.
+     * @param languageId - The uuid of the language.
+     */
+    async assignSalesChannelLanguage(salesChannelId: string, languageId: string) {
```
frobel, line 1243: I would go here not with a patch, because then we could not make it back and stay the whole test execution with the assigned language within the Sales Channel. That would affect other test cases. I would go here with a sync API command for the table sales-channel-language...here a snippet. 

```
const syncSalesChannelResponse = await this.AdminApiClient.post('./_action/sync', {
            data: {
                'write-sales-channel-language': {
                    entity: 'sales_channel_language',
                    action: 'upsert',
                    payload: [
                        {
                            salesChannelId: salesChannelId,
                            languageId: languageId,
                        },
                    ],
                },
            },
        });
        expect(syncSalesChannelResponse.ok()).toBeTruthy();

        const { data: salesChannel } = await syncSalesChannelResponse.json();

        this.addCreatedRecord('sales_channel_language', {salesChannelId: salesChannelId, languageId: languageId})

    private highPriorityEntities = ['order', 'product', 'landing_page', 'sales_channel_currency', 'sales_channel_country', 'sales_channel_language'];

```

frobel: see here: https://github.com/shopware/acceptance-test-suite/blob/6a0982898de60b4b118abff2b61b80d82a5bce2b/src/services/TestDataService.ts#L1101

vanpham-sw: Updated. I need to delete customer before the **sales_channel_language**, if not the customers those linked to the added language won't be deleted.

frobel: mhh i will double check

frobel: Okay...I double check it and your current implementation is doing what it should. I assign a language to the sales channel (available in the sales channel settings) and also delete it properly. But if you add a custom language and assign it to your sales channel you have to add to the priority list: `sales_channel_language` ...because that entry has to be deleted first before the `language` entity is deleted

frobel: Regarding your problem with the customer deletion it is another one and not related to that thread.

src/services/TestDataService.ts
```diff
@@ -97,7 +100,7 @@ export class TestDataService {
      *
      * @private
      */
-    private highPriorityEntities = ['order', 'product', 'landing_page', 'sales_channel_currency', 'sales_channel_country'];
+    private highPriorityEntities = ['order', 'product', 'landing_page', 'sales_channel_currency', 'sales_channel_country', 'sales_channel_language'];
```
frobel: I figured out that we have to add 'customer' here and delete 'sales_channel_language' because during the deletion of a customer with a assigned language, the language has to be present in the sales channel as long the customer is available with that specific language.

frobel (approved):

PR #196, opened 2024-11-29
feat: system config service in TestDataService and page object improv…

frobel:
* system config service in TestDataService 
* page object improvements

### DIFF
--- a/src/page-objects/storefront/AccountOrder.ts
+++ b/src/page-objects/storefront/AccountOrder.ts
@@ -6,12 +6,55 @@ export class AccountOrder implements PageObject {
     public readonly cartLineItemImages: Locator;
     public readonly orderExpandButton: Locator;
     public readonly digitalProductDownloadButton: Locator;
+    public readonly dialogOrderCancel: Locator;
+    public readonly dialogOrderCancelButton: Locator;
+    public readonly dialogBackButton: Locator;
 
     constructor(public readonly page: Page) {
         this.orderExpandButton = page.getByRole('button', {name: /Expand|Show details/}).first();
         this.cartLineItemImages = page.locator('.line-item-img-link');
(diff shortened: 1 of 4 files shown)


src/page-objects/storefront/AccountOrder.ts
```diff
@@ -6,12 +6,55 @@ export class AccountOrder implements PageObject {
...
+        const orderStatus = orderItem.locator('.order-table-header-order-status');
+        const orderHeading = orderItem.locator('.order-table-header-heading');
+        const orderActionsButton = orderItem.getByLabel('Actions');
+        const orderCancelButton = orderItem.getByRole('button', { name: 'Cancel order' });
+        const orderRepeatButton = orderItem.getByRole('button', { name: 'Repeat order' });
+        const orderChangePaymentMethodButton = orderItem.getByRole('link', { name: 'Change payment method' });
+        const orderShippingStatus = orderItem.getByLabel('Shipping status: ');
+        const orderPaymentStatus = orderItem.getByLabel('Payment status: ');
+        const orderPaymentMethod = orderItem.getByLabel('Payment method: ');
+        const orderShippingMethod = orderItem.getByLabel('Shipping method: ');
```
yusufttur: Why do you have spaces at the end?

yusufttur (dismissed):

src/page-objects/storefront/AccountOrder.ts
frobel: it is defined in the platform code. Maybe it works also without it.

frobel: I checked again and the A11Y is not so proper and you found a bug. Therefore I need to use the `nth()` method.

yusufttur (approved):

PR #198, opened 2024-12-02
NEXT-39662 - rule assignment test

GitEvil:
adds task and page objects for rule assignment test


src/tasks/shop-admin/Rule/AssignEntityToRule.ts
```diff
@@ -0,0 +1,70 @@
...
+                    await AdminRuleDetail.promotionCustomerRuleCard.getByRole('link').filter({hasText: entity.name}).click();
+                }
+                if (entity.entityType == 'promotionCartRule') {
+                    const promotionCartRuleCard = await AdminRuleDetail.getRuleCardContent(AdminRuleDetail.promotionCartRuleCard);
+                    await promotionCartRuleCard.addAssignmentButton.click();
+                    await AdminRuleDetail.page.locator('.sw-settings-rule-add-assignment-modal').getByPlaceholder('Search...').fill(entity.name);
+                    await AdminRuleDetail.page.locator('.sw-data-grid__row').filter({hasText: entity.name}).getByRole('checkbox').click();
+                    await AdminRuleDetail.page.locator('.sw-button--primary').getByText('Add').click();
+                    await AdminRuleDetail.promotionCartRuleCard.getByRole('link').filter({hasText: entity.name}).click();
+                }
```
vanpham-sw:
```suggestion
                async function handleEntityAssignment(entityType, card) {
    const cardContent = await AdminRuleDetail.getRuleCardContent(card);
    await cardContent.addAssignmentButton.click();
    await AdminRuleDetail.page.locator('.sw-settings-rule-add-assignment-modal').getByPlaceholder('Search...').fill(entity.name);
    await AdminRuleDetail.page.locator('.sw-data-grid__row').filter({ hasText: entity.name }).getByRole('checkbox').click();
    await AdminRuleDetail.page.locator('.sw-button--primary').getByText('Add').click();
    await cardContent.getByRole('link').filter({ hasText: entity.name }).click();
}

switch (entity.entityType) {
    case 'shippingMethod':
        await handleEntityAssignment(entity.entityType, AdminRuleDetail.shippingMethodCard);
        break;
    case 'taxProvider':
        await handleEntityAssignment(entity.entityType, AdminRuleDetail.taxProviderCard);
        break;
    case 'paymentMethod':
        await handleEntityAssignment(entity.entityType, AdminRuleDetail.paymentMethodCard);
        break;
    case 'promotionOrderRule':
        await handleEntityAssignment(entity.entityType, AdminRuleDetail.promotionOrderRuleCard);
        break;
    case 'promotionCustomerRule':
        await handleEntityAssignment(entity.entityType, AdminRuleDetail.promotionCustomerRuleCard);
        break;
    case 'promotionCartRule':
        await handleEntityAssignment(entity.entityType, AdminRuleDetail.promotionCartRuleCard);
        break;
    default:
        throw new Error(`Unknown entity type: ${entity.entityType}`);
}
```

src/tasks/shop-admin/Rule/AssignEntityToRule.ts
```diff
@@ -0,0 +1,70 @@
...
+import type {FixtureTypes} from '../../../types/FixtureTypes';
+import {Rule} from '../../../types/ShopwareTypes';
+
+interface Entity {
+    id: string;
+    name: string;
+    entityType: string;
+// 'shippingMethod' | 'taxProvider' | 'paymentMethod' | 'promotionOrderRule' | 'promotionCustomerRule' | 'promotionCartRule';
+}
+export const AssignEntityToRule = base.extend<{ AssignEntityToRule: Task }, FixtureTypes>({
```
jc-sw-git: The method does more, than the name implies. Clicking on the entity to open its detail page should not be part of the assignment task. Instead it should be a separate action in your test, which starts basically the validation to check, if the correct entity has been assigned.

src/page-objects/administration/RuleDetail.ts
```diff
@@ -3,21 +3,43 @@ import type { PageObject } from '../../types/PageObject';
...
+    public readonly promotionOrderRuleCard: Locator;
+    public readonly promotionCustomerRuleCard: Locator;
+    public readonly promotionCartRuleCard: Locator;
+
     constructor(public readonly page: Page) {
         this.nameInput = page.getByLabel('Name');
         this.priorityInput = page.getByLabel('Priority');
         this.conditionTypeSelectionInput = page.locator('.sw-condition-type-select').locator('.sw-single-select__selection');
         this.conditionValueSelectionInput = page.locator('.sw-condition__value-content').locator('.sw-entity-single-select__selection');
         this.filtersResultPopoverSelectionList = page.locator('.sw-select-result-list__content').getByRole('listitem');
```
jc-sw-git: We also have a RuleCreate page object, right? I would expect that they have the same locators here? Would you mind to refactor this page object, when you are already touching it, so that the RuleDetail is basically just an extension of the RuleCreate and only adds the locators, which are exclusive to the detail page (the assignment tab) but can still make use of the locators, which are also available in RuleCreate?
That way, we would only need to adjust the locators in RuleCreate, if they would for some reason change, and they would automatically be adjusted for RuleDetail as well.

GitEvil:
/

frobel (dismissed):

src/page-objects/administration/PaymentDetail.ts
```diff
@@ -0,0 +1,20 @@
...
+export class PaymentDetail implements PageObject {
+
+    public readonly header: Locator;
+    public readonly nameField: Locator;
+    public readonly availabilityRule: Locator;
+
+
+    constructor(public readonly page: Page) {
+        this.header = page.locator('.smart-bar__header');
+        this.nameField = page.getByLabel('Name').first();
```
frobel: Why you took first() here? Is there a second one? if yes, its possible to make it unique somehow?

src/page-objects/administration/PaymentDetail.ts
```diff
@@ -0,0 +1,20 @@
...
+        this.availabilityRule = page.locator('.sw-settings-payment-detail__field-availability-rule').locator('.sw-entity-single-select__selection-text');
```
frobel: mhh it is better to use here the following approach or?
```
this.ruleSelectionInput = page.locator('..sw-settings-payment-detail__field-availability-rule').locator('.sw-entity-single-select__selection');
        this.filtersResultPopoverSelectionList = page.locator('.sw-select-result-list__content').getByRole('listitem');
```

src/page-objects/administration/RuleListing.ts
```diff
@@ -0,0 +1,16 @@
...
+
+    public readonly createRuleButton: Locator;
+
+    constructor(public readonly page: Page) {
+        this.createRuleButton = page.getByText('Create rule');
+
+    }
+
+    url() {
+        return `#/sw/settings/rule/index`
```
frobel:
```suggestion
        return `#/sw/settings/rule/index`;
```

src/page-objects/administration/RuleDetail.ts
```diff
@@ -1,23 +1,45 @@
...
+    public readonly promotionCustomerRulesCard: Locator;
+    public readonly promotionCartRulesCard: Locator;
 
     constructor(public readonly page: Page) {
-        this.nameInput = page.getByLabel('Name');
-        this.priorityInput = page.getByLabel('Priority');
-        this.conditionTypeSelectionInput = page.locator('.sw-condition-type-select').locator('.sw-single-select__selection');
-        this.conditionValueSelectionInput = page.locator('.sw-condition__value-content').locator('.sw-entity-single-select__selection');
-        this.filtersResultPopoverSelectionList = page.locator('.sw-select-result-list__content').getByRole('listitem');
+        super(page)
```
frobel:
```suggestion
        super(page);
```

frobel (changes requested):

src/page-objects/administration/PaymentDetail.ts
GitEvil: The second one is the technical name field. But you were right, there is a better locator to pinpoint the actual name field. 👌

src/page-objects/administration/PaymentDetail.ts
GitEvil: Well, I don't think this improves the test much, but it sure was a nice coding exercise. However, it will add quite a lot of code for—if you ask me—no good reason. What do you think?

src/page-objects/administration/PaymentDetail.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    public readonly header: Locator;
+    public readonly nameField: Locator;
+    public readonly availabilityRuleField: Locator;
+    public readonly availabilityRuleListItem: Locator;
+
+
+    constructor(public readonly page: Page) {
+        this.header = page.locator('.smart-bar__header');
+        this.nameField = page.locator('.sw-settings-payment-detail__field-name').getByLabel('Name');
```
frobel: I double checked. The problem here why you get 2 results with getByLabel('Name'), is getByLabel checks where a label is present which contains "name" ...in your case of course Name field and technical name field because the label of technical name field also called "technical _name_" 

What you could do here to use the locator with "exact: true"

page.getByLabel('Name', { exact: true });

frobel (dismissed):

src/tasks/shop-admin/Rule/AssignEntitiesToRule.ts
```diff
@@ -0,0 +1,57 @@
+import {Locator, test as base} from '@playwright/test';
+import type {Task} from '../../../types/Task';
+import type {FixtureTypes} from '../../../types/FixtureTypes';
+
+interface Entity {
+    id: string;
+    name: string;
+    ruleType: string;
+}
```
jc-sw-git: Is there a way to use the actual entity as a parameter? That way, the dev, who wants to use this method, wouldn't need to strip down the entitiy and define the ruleType per entry themself. Instead it would be nice, if the method could identify the ruleType based on the entities, that handed over by the parameter.

src/tasks/shop-admin/Rule/AssignEntitiesToRule.ts
```diff
@@ -0,0 +1,57 @@
...
+                }
+                for (const entity of assignableEntities) {
+                    const validRuleType = {
+                        shipping1: 'shippingMethodAvailabilityRule',
+                        tax1: 'taxProviderRule',
+                        payment1: 'paymentMethodAvailabilityRule',
+                        promotion1: 'promotionOrderRule',
+                        promotion2: 'promotionCustomerRule',
+                        promotion3: 'promotionCartRule',
+                    };
```
jc-sw-git: Why is it necessary to assign the const each time in the loop? Why do we even need the consts at all? Couldn't you just use the values directly in the case statements? And if it is necessary, I would probably use better understandable naming like shippingMethod, taxProvider, promotionOrderRule instead of shipping1, tax1, promotion1 etc.

frobel: valid point

src/tasks/shop-admin/Rule/AssignEntitiesToRule.ts
frobel: we have also here a general problem, that the task currently is not providing any information in form of a documentation what the Task need/want as parameter...but yes what kind of ruleTypes are possible is not visible here. We have also other tests where parameters are not 100% sure how to set.

jc-sw-git: @frobel regarding the general problem, we can probably talk about it a bit more in our 1:1

frobel (approved):

PR #225, opened 2024-12-11
fix!: fix saas related issues

pweyck:


src/page-objects/administration/Categories.ts
```diff
@@ -14,7 +14,15 @@ export class Categories implements PageObject {
...
+
+  /**
+   * @deprecated Please don't depend on the default categories
+   */
   public readonly homeCategory: Locator;
+
+  /**
+   * @deprecated Please don't depend on the default categories
+   */
   public readonly homeCategoryContextButton: Locator;
```
pweyck: Maybe it's better to remove this and break it, because we don't do this anyway

pweyck: I've removed the home stuff

src/page-objects/administration/Categories.ts
```diff
@@ -14,8 +14,6 @@ export class Categories implements PageObject {
    * Category tree
    */
   public readonly categoryTree: Locator;
-  public readonly homeCategory: Locator;
```
pweyck, line 17: saas and potentially many other systems do not have the home category. We should always create all data that we expect to be in the system

vanpham-sw (approved):

src/page-objects/administration/Categories.ts
```diff
@@ -51,8 +49,7 @@ export class Categories implements PageObject {
     this.addLandingPageButton = this.landingPageArea.getByText('Add landing page');
     this.landingPageItems = this.landingPageArea.locator('.sw-tree-item__label');
     this.categoryTree = page.locator('.sw-category-tree');
-    this.homeCategory = this.categoryTree.locator('.sw-tree-item__element').filter({ hasText: 'Home' });
```
frobel, line 54: if we remove that, we have to adjust the a platform test or?

tests/TestDataService/TestDataService.spec.ts
```diff
@@ -143,9 +141,16 @@ test('Data Service', async ({
     const { data: databaseCategory } = (await categoryResponse.json()) as { data: Category };
     expect(databaseCategory.id).toBe(category.id);
 
-    const systemConfigEntryResponse = await AdminApiContext.get(`./system-config/${systemConfigEntry.id}?_response=detail`);
-    const { data: databaseSystemConfigEntry } = (await systemConfigEntryResponse.json()) as { data: SystemConfig };
-    expect(databaseSystemConfigEntry.id).toBe(systemConfigEntry.id);
+    // eslint-disable-next-line playwright/no-conditional-in-test
+    if (!isSaaSInstance(AdminApiContext)) {
```
frobel, line 145: ow yes, good point

frobel (approved):

src/page-objects/administration/Categories.ts
frobel: @vanpham-sw you aware of it?

vanpham-sw: Yes, we have to adjust the platform test:
From this:
`await ShopAdmin.attemptsTo(CreateLinkTypeCategory(categoryData, categoryCustomizableLinkData));`
to:
`await ShopAdmin.attemptsTo(CreateLinkTypeCategory(categoryData, categoryCustomizableLinkData, 'Home'));`

I can create a MR, but maybe there is a better way?

pweyck: I plan to do that :)

PR #228, opened 2024-12-12
fix: type check and improve CreateLinkTypeCategory task

vanpham-sw:


src/tasks/shop-admin/Category/CreateLinkTypeCategory.ts
```diff
@@ -2,10 +2,25 @@ import { test as base, Locator } from '@playwright/test';
 import type { Task } from '../../../types/Task';
 import type { FixtureTypes } from '../../../types/FixtureTypes';
 
+interface CategoryData {
+  name: string;
+  categoryType: string;
```
frobel: what about that type?

frobel (changes requested):

frobel (approved):

PR #230, opened 2024-12-12
fix: use correct system-config endpoint

pweyck:


src/services/TestDataService.ts
```diff
@@ -875,27 +875,16 @@ export class TestDataService {
...
-            configurationValue: configurationValue,
-            salesChannelId: salesChannelId || null,
-        }
+    ): Promise<void> {
+        const data: Record<string, unknown> = {};
+        data[configurationKey] = configurationValue;
 
-        const response = await this.AdminApiClient.post('system-config?_response=detail', {
-            data: systemConfigStruct,
+        const response = await this.AdminApiClient.post(`_action/system-config?_response=detail&salesChanneldId=${salesChannelId}`, {
```
frobel: Somehow we have to delete the systemConfig entry here. Should be built-in that into the test case?

pweyck: I'd be pretty complicated. If you set a config you don't actually know if it's already set. So you would need to know what the previous value was save it and then restore it in the end

pweyck: to be honest, I'd only allow setting it per sales channel basis and just unset all sales channel specific values at the end

pweyck: I've changed it to unset all sales channel specific values. The api is also changed, so that it only allow settings sales channel specific values

src/tasks/shop-admin/Settings/SetSystemConfigValues.ts
```diff
@@ -1,34 +1,15 @@
 import { test as base, expect } from '@playwright/test';
 import type { Task } from '../../../types/Task';
-import type { FixtureTypes} from '../../../types/FixtureTypes';
+import type { FixtureTypes } from '../../../types/FixtureTypes';
 
 export const SetSystemConfigValues = base.extend<{ SetSystemConfigValues: Task }, FixtureTypes>({
```
frobel: Do we need actually that task. We could directly use the TestDataService method `setSystemConfig()` in our tests.

src/services/TestDataService.ts
```diff
@@ -865,37 +867,23 @@ export class TestDataService {
...
         });
         expect(response.ok()).toBeTruthy();
 
-        const { data: systemConfigEntry } = (await response.json()) as { data: SystemConfig };
-
-        this.addCreatedRecord('system_config', systemConfigEntry.id);
+        // we just unset all configs with this.defaultSalesChannel.id
+        // the defaults should take effect again
+        for (const key of Object.keys(configs)) {
+            this.restoreSystemConfig[key] = null;
```
frobel, line 884: Where the restore is happening here? I just see that the values are set to zero but the database still have the newValues added by the method `setSystemConfig(configs)`. Or I interpreted the API Call wrong?

frobel (changes requested):

src/tasks/shop-admin/Settings/SetSystemConfigValues.ts
pweyck: I've removed it

src/services/TestDataService.ts
pweyck: https://github.com/shopware/acceptance-test-suite/pull/230/files#diff-a733b20ef0fec95da98ad2e7f38bb05601a59f1597f7fbfad202824c12dca4f5R1704-R1708

frobel (dismissed):

PR #237, opened 2024-12-16
Refactor Rule assignment task and types

GitEvil:


src/types/ShopwareTypes.ts
```diff
@@ -242,9 +242,12 @@ export enum RuleType {
     promotionCustomer = 'promotionCustomerRule',
     promotionCart = 'promotionCartRule',
 }
-export interface RuleAssignmentEntity {
+
+export interface RuleAssignmentEntities {
```
frobel: Why do you define it as Entities? Usually it is defining a single entity and as soon as you add an [] Array to it the developer knows okay it is more then one.

frobel: And why you exclude the ruleType from the entity?

frobel (changes requested):

src/tasks/shop-admin/Rule/AssignEntitiesToRule.ts
```diff
@@ -2,11 +2,11 @@ import {Locator, test as base} from '@playwright/test';
 import type {Task} from '../../../types/Task';
 import type {FixtureTypes} from '../../../types/FixtureTypes';
 import {RuleType} from '../../../types/ShopwareTypes';
```
frobel:
```suggestion
import type { Task } from '../../../types/Task';
import type { FixtureTypes } from '../../../types/FixtureTypes';
import { RuleType } from '../../../types/ShopwareTypes';
```

src/tasks/shop-admin/Rule/AssignEntitiesToRule.ts
```diff
@@ -2,11 +2,11 @@ import {Locator, test as base} from '@playwright/test';
...
-import {RuleAssignmentEntity} from '../../../types/ShopwareTypes';
+import {RuleAssignmentEntities} from '../../../types/ShopwareTypes';
```
frobel:
```suggestion
import { RuleAssignmentEntities } from '../../../types/ShopwareTypes';
```

frobel (changes requested):

src/types/ShopwareTypes.ts
GitEvil: The ruleType cannot be extracted from the Entity created by the TestDataService but has to be decided by the test, so it is not part of the entity object.

frobel (approved):

PR #240, opened 2024-12-17
fix: NEXT-39784 - fix and add some locators to AccountProfile and Acc…

ocavli:
needed to adjust some locators and add some news to the pages.

### DIFF
--- a/src/page-objects/storefront/AccountProfile.ts
+++ b/src/page-objects/storefront/AccountProfile.ts
@@ -22,6 +22,11 @@ export class AccountProfile implements PageObject {
 
     public readonly emailUpdateMessage: Locator;
     public readonly passwordUpdateMessage: Locator;
+    public readonly emailValidationAlert: Locator;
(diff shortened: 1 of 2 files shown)


src/page-objects/storefront/AccountProfile.ts
```diff
@@ -22,6 +22,11 @@ export class AccountProfile implements PageObject {
 
     public readonly emailUpdateMessage: Locator;
     public readonly passwordUpdateMessage: Locator;
+    public readonly emailValidationAlert: Locator;
+    public readonly emailUpdateFailureAlert: Locator;
+    public readonly passwordUpdateFailureAlert: Locator;
+
+
```
frobel:
```suggestion
```

frobel (changes requested):

src/page-objects/storefront/AccountProfile.ts
ocavli: fixed 👍

frobel (approved):

PR #244, opened 2024-12-18
Next 39785/customer get access back when forgot password

vanpham-sw:


src/page-objects/storefront/AccountRecover.ts
```diff
@@ -19,6 +22,9 @@ export class AccountRecover implements PageObject {
         this.requestEmailButton = this.passwordRecoveryForm.getByRole('button', { name: 'Request email' });
         this.backButton = this.passwordRecoveryForm.getByRole('link', { name: 'Back' });
         this.passwordResetEmailSentMessage = page.getByText('If the provided email address is registered, a confirmation email including a password reset link has been sent.');
+        this.newPasswordInput = page.getByLabel('New password');
+        this.newPasswordConfirmInput = page.getByLabel('Password confirmation');
+        this.changePasswordButton = page.getByRole('button', { name: 'Change password' });
     }
 
     url() {
```
yusufttur:
```suggestion
    url(?recoveryLink: string) {
          if(recoveryLink){
                return recoveryLink;
                }
```
Would you like to add an optional string parameter to add your link directly here?
`await ShopCustomer.goesTo(StorefrontAccountRecover.url(await MailpitApiContext.getLinkFromMail(customer.email)));`

vanpham-sw: yes, updated @yusufttur thanks

yusufttur (approved):

PR #247, opened 2024-12-19
feat: add features data to InstanceMeta fixture

pweyck:


src/fixtures/HelperFixtures.ts
```diff
@@ -61,11 +65,22 @@ export const test = base.extend<NonNullable<unknown>, FixtureTypes>({
         async ({ AdminApiContext: context }, use) => {
             const response = await context.get('./_info/config');
             expect(response.ok(), '/_info/config request failed').toBeTruthy();
-
             const config = (await response.json()) as { version: string };
+
+            const features: FeaturesType = {};
+            if (satisfies(config.version, '>=6.6.0.0')) {
+                const featuresResponse = await context.get('./_action/feature-flag');
```
philipreinken, line 72: Is this version requirement chosen based on the API endpoint only being available from 6.6 onwards?

philipreinken (approved):

frobel (approved):

src/fixtures/HelperFixtures.ts
pweyck: yes, the endpoint was added in 6.6

PR #249, opened 2024-12-20
fix: example for locator comatibility in shopware versions

frobel:


src/page-objects/administration/CustomerListing.ts
```diff
@@ -13,10 +15,16 @@ export class CustomerListing implements PageObject {
...
     public readonly modalHeaderCheckbox: Locator;
 
-    constructor(public readonly page: Page) {
+    constructor(public readonly page: Page, public readonly instanceMeta: HelperFixtureTypes['InstanceMeta']) {
         this.headline = page.getByRole('heading', { name: 'Customers' });
         this.addCustomerButton = page.locator('.sw-customer-list__button-create');
         this.bulkEditButton = page.getByRole('button', { name: 'Bulk edit' });
+
+        // Locator available in versions < 6.7
+        if(satisfies(instanceMeta.version, '<6.7')){
```
frobel, line 24: For platform PRs (running test pipeline with major feature flag) there is another possibility:

https://github.com/shopware/shopware/pull/5973

src/page-objects/administration/CustomerListing.ts
```diff
@@ -13,11 +15,18 @@ export class CustomerListing implements PageObject {
...
```
yusufttur, line 24: @frobel This doesn't work for 6.7 instances, even major feature flag is activated, it returns `6.6.9999999.9999999-dev` as version number in the console.

yusufttur (changes requested):

src/page-objects/administration/CustomerListing.ts
frobel: yes true, @pweyck want to provide a fix so that we get back 6.7.999.999-dev

pweyck: https://github.com/shopware/shopware/pull/5993

PR #257, opened 2025-01-06
fix: add additional locator for smtp input fields within FRW

frobel:


ocavli (approved):

src/page-objects/administration/FirstRunWizard.ts
```diff
@@ -29,8 +29,21 @@ export class FirstRunWizard implements PageObject {
     public readonly salesChannelSelectionList: Locator;
     public readonly salesChannelSelectionMultiSelect: Locator;
     public readonly smtpServerTitle: Locator;
+
+    /**
+     * @deprecated - Use `smtpServerFieldInputs` instead.
```
yusufttur, line 34: Which locators are deprecated? I guess we need to add 6.8 tag or something similar to not to forget to remove it when 6.8 were introduced.

yusufttur (changes requested):

src/page-objects/administration/FirstRunWizard.ts
frobel: it is just a deprecation for us so if we make a new acceptance test suite major version.

frobel: it is not shopware platform related.

PR #261, opened 2025-01-09
fix: locator compatibility of product image

vanpham-sw:


src/page-objects/storefront/Home.ts
```diff
@@ -30,7 +30,7 @@ export class Home implements PageObject {
         if (satisfies(instanceMeta.version, '<6.7')) {
             this.productImages = page.locator('.product-image-link');
         } else {
-            this.productImages = page.locator('.product-image');
+            this.productImages = page.locator('.product-image-wrapper');
```
yusufttur: I think this locator is available in both versions, I recommend you to either use .product-image as I used before or use this one without any if clause.

yusufttur (changes requested):

src/page-objects/storefront/Home.ts
vanpham-sw: When using .product_image it fails to get the element **getByAltText** on this line: StorefrontHome.productImages.getByAltText(media.alt)
Because the .product-image is the last node. So I'm using .product-image-wrapper

yusufttur (approved):

PR #266, opened 2025-01-16
feat!: change register task for guest users

yusufttur:
I would like to use register task for guest & commercial users

### DIFF
--- a/src/tasks/shop-customer/Account/Register.ts
+++ b/src/tasks/shop-customer/Account/Register.ts
@@ -26,13 +26,13 @@ export const Register = base.extend<{ Register: Task }, FixtureTypes>({
             vatRegNo: 'DE1234567890',
         };
 
-        const task = (overrides?: Partial<RegistrationData>) => {
+        const task = (overrides?: Partial<RegistrationData>, isCommercial?: boolean) => {
             return async function Register() {
                 const registrationData = { ...defaultRegistrationData, ...overrides };
 
                 registeredEmail = registrationData.email;
 
-                if (registrationData.isCommercial) {
+                if (registrationData.isCommercial || isCommercial) {
                     await StorefrontAccountLogin.accountTypeSelect.selectOption('Commercial');
                     await StorefrontAccountLogin.companyInput.fill(registrationData.company);
                     await StorefrontAccountLogin.departmentInput.fill(registrationData.department);
--- a/src/tasks/shop-customer/Account/Register.ts
+++ b/src/tasks/shop-customer/Account/Register.ts
@@ -5,17 +5,13 @@ import type { components } from '@shopware/api-client/admin-api-types';
 import type { RegistrationData } from '../../../types/ShopwareTypes';
 import { AdminApiContext } from 'src/services/AdminApiContext';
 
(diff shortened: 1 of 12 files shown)


src/tasks/shop-customer/Account/Register.ts
```diff
@@ -1,10 +1,16 @@
 import { test as base } from '@playwright/test';
 import type { Task } from '../../../types/Task';
-import type { FixtureTypes} from '../../../types/FixtureTypes';
+import type { FixtureTypes } from '../../../types/FixtureTypes';
 import type { components } from '@shopware/api-client/admin-api-types';
 import type { RegistrationData } from '../../../types/ShopwareTypes';
 import { AdminApiContext } from 'src/services/AdminApiContext';
 
+interface RegistrationOptions {
```
vanpham-sw: Should we move RegistrationOptions to ShopwareTypes?

vanpham-sw (dismissed):

src/tasks/shop-customer/Account/Register.ts
yusufttur: I don't think so, it is special to this task.

src/tasks/shop-customer/Account/Register.ts
```diff
@@ -13,71 +19,85 @@ export const Register = base.extend<{ Register: Task }, FixtureTypes>({
...
             country: 'Germany',
             postalCode: '48624',
-            company: 'shopware', 
-            department: 'Operations', 
+            company: 'shopware',
+            department: 'Operations',
             vatRegNo: 'DE1234567890',
         };
 
-        const registerTask = (overrides?: Partial<RegistrationData>, isCommercial?: boolean ) => {
```
frobel, line 27: Why do you change here to `options` and not just add `isGuest?: boolean` to the method?

yusufttur: I could do it like that, but then readability in the spec files was not good. Which one do you prefer?
```
await ShopCustomer.attemptsTo(Register( customer, true, true ));
await ShopCustomer.attemptsTo(Register({ overrides: customer, isCommercial: true, isGuest: true }));
```

frobel: I like more the second one, but lets go with the first approach (customer, true, true) and think about a solution how we can enable the code suggestion for tasks and refactor it in a new major for all tasks

frobel: And on the other hand your change is not creating a breaking change if you add isGuest as a new parameter in the method on the last position. But please don't forget to deprecate the task RegisterGuest(). thx

yusufttur: I added isGuest and isCommercial into registrationData type so it shouldn't be a big change. What do you think? And this is how it looks like in the test:
 `await ShopCustomer.attemptsTo(Register());`
 `await ShopCustomer.attemptsTo(Register(customer));`
 `await ShopCustomer.attemptsTo(Register({ isGuest: true }));`
 `await ShopCustomer.attemptsTo(Register({ isCommercial: true, isGuest: true }));`

PR #276, opened 2025-01-20
fix: add additional locator for main navigation

frobel:


src/page-objects/storefront/Home.ts
```diff
@@ -2,9 +2,6 @@ import type { Page, Locator } from '@playwright/test';
 import type { PageObject } from '../../types/PageObject';
 
 export class Home implements PageObject {
-
-    public readonly accountMenuButton: Locator;
-    public readonly closeGuestSessionButton: Locator;
```
yusufttur: Why you're deleting these locators? I need them in my guest customer register test. 🤔

yusufttur (changes requested):

src/page-objects/storefront/Home.ts
frobel: Sorry it was not intended. Thx to mention it...I see there are also other locators which gets deleted.

yusufttur (approved):

PR #280, opened 2025-01-20
fix: NEXT-40281 - added some new locators and method to search suggest

ocavli:
To create search test, need to add some locators and method on search suggest page.

### DIFF
--- a/src/page-objects/storefront/SearchSuggest.ts
+++ b/src/page-objects/storefront/SearchSuggest.ts
@@ -4,9 +4,48 @@ import type { PageObject } from '../../types/PageObject';
 export class SearchSuggest implements PageObject {
 
     public readonly searchSuggestLineItemImages: Locator;
+    public readonly searchInput: Locator;
+    public readonly searchIcon: Locator;
+    public readonly searchSuggestNoResult: Locator;
+    public readonly searchSuggestLineItemName: Locator;
+    public readonly searchSuggestLineItemPrice: Locator;
+    public readonly searchSuggestTotalLink: Locator;
+    public readonly searchHeadline: Locator;
 
     constructor(public readonly page: Page) {
         this.searchSuggestLineItemImages = page.locator('.search-suggest-product-image-container');
+        this.searchInput = page.locator('.header-search-input');
+        this.searchIcon = page.locator('.header-search-icon');
+        this.searchSuggestNoResult = page.locator ('.search-suggest-no-result');
+        this.searchSuggestLineItemName = page.locator ('.search-suggest-product-name');
+        this.searchSuggestLineItemPrice = page.locator ('.col-auto.search-suggest-product-price');
+        this.searchSuggestTotalLink = page.locator ('.search-suggest-total-link');
+        this.searchHeadline = page.locator ('.search-headline');
+    }
+
+    async getListingItemByProductId(productId: string): Promise<Record<string, Locator>> {
+        const listingItem = this.page.getByRole('listitem').filter({ has: this.page.locator(`[value="${productId}"]`) });
+        const productImage = listingItem.locator('.product-image-link');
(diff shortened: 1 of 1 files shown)


src/page-objects/storefront/SearchSuggest.ts
```diff
@@ -4,9 +4,48 @@ import type { PageObject } from '../../types/PageObject';
...
+            productRating: productRating,
+            productVariantCharacteristics: productVariantCharacteristics,
+            productDescription: productDescription,
+            productPriceUnit: productPriceUnit,
+            productCheapestPrice: productCheapestPrice,
+            productPrice: productPrice,
+            productName: productName,
+            productAddToShoppingCart: productAddToShoppingCart,
+        }
     }
```
yusufttur: You can implement this by extending this page from Home.ts, so no need to rewrite it again.

yusufttur (changes requested):

src/page-objects/storefront/SearchSuggest.ts
ocavli: added your feedback 👍

src/page-objects/storefront/SearchSuggest.ts
```diff
@@ -1,15 +1,31 @@
...
+        this.searchSuggestNoResult = page.locator ('.search-suggest-no-result');
+        this.searchSuggestLineItemName = page.locator ('.search-suggest-product-name');
+        this.searchSuggestLineItemPrice = page.locator ('.col-auto.search-suggest-product-price');
+        this.searchSuggestTotalLink = page.locator ('.search-suggest-total-link');
+        this.searchHeadline = page.locator ('.search-headline');
     }
 
-    url(searchTerm: string) {
+    url(searchTerm?: string) {
         return `suggest?search=${searchTerm}`;
```
yusufttur, line 29: It should be possible to land this page with and without searchTerm

yusufttur (changes requested):

src/page-objects/storefront/SearchSuggest.ts
```diff
@@ -1,15 +1,31 @@
 import type { Page, Locator } from '@playwright/test';
 import type { PageObject } from '../../types/PageObject';
+import { Home } from './Home';
 
-export class SearchSuggest implements PageObject {
+export class SearchSuggest extends Home implements PageObject {
```
frobel, line 5: Why you are extending the Class?

frobel (changes requested):

src/page-objects/storefront/SearchSuggest.ts
yusufttur: https://github.com/shopware/acceptance-test-suite/pull/280#discussion_r1923201159

ocavli: It was Yusuf's first feedback to use the method directly there. Also, the page is similar to the home.
Is there any way to use just the method from another page without extending it? Or do you have another suggestion?

yusufttur (approved):

src/page-objects/storefront/SearchSuggest.ts
frobel: What do you think about to use the `StorefrontHome` fixture and the `StorefrontSuggestSearch` fixture in your platform test instead of just using `StorefrontSuggestSearch` in your test.

This is just an idea. But in general, your solution is correct because you extend the Suggest Search fixture with the HOME fixture properties.

frobel (approved):

src/page-objects/storefront/SearchSuggest.ts
ocavli: The elements and also usage of the method are the similar as HOME, so I would go with this solution for now. but it might be an idea for future tests.

PR #281, opened 2025-01-21
feat: add tax rate creation method via TestDataService

frobel:


src/services/TestDataService.ts
```diff
@@ -983,6 +983,27 @@ export class TestDataService {
...
     }
 
+    /**
+     * Creates a new tax rate (19%) with a random name.
+     *
+     * @param overrides - Specific data overrides that will be applied to the tax data struct.
+     */
+    async createTaxRate(
+        overrides: Partial<Tax> = {},
+    ): Promise<Tax> {
```
yusufttur: I think prettier brokes our files. This could be a one line. 😄

src/services/TestDataService.ts
```diff
@@ -31,7 +31,7 @@ import type {
     SalesChannelDomain,
     Language,
     CustomFieldSet,
-    CustomField,
+    CustomField, Tax,
```
yusufttur: Please move it to next line.

src/services/TestDataService.ts
```diff
@@ -2569,4 +2590,18 @@ export class TestDataService {
 
         return Object.assign({}, basicSalesChannelDomain, overrides);
     }
+
+    getTaxStruct(overrides: Partial<Tax>): Partial<Tax> {
+
```
yusufttur: remove empty line please

tests/TestDataService/TestDataService.spec.ts
```diff
@@ -142,6 +146,10 @@ test('Data Service', async ({
     const { data: databaseSalesChannelAnalytics } = (await salesChannelAnalyticsResponse.json()) as { data: SalesChannelAnalytics };
     expect(databaseSalesChannelAnalytics.id).toBe(salesChannelAnalytics.id);
 
+    const taxRate21Response = await AdminApiContext.get(`./tax/${taxRate21.id}?_response=detail`);
+    const { data: databaseTaxRate21 } = (await taxRate21Response.json()) as { data: Tax };
+    expect(databaseTaxRate21.id).toBe(taxRate21.id);
```
yusufttur, line 151: Is it something new? Interesting, but good that we're not only checking method but also API call. 💯

yusufttur (changes requested):

tests/TestDataService/TestDataService.spec.ts
frobel: It is not new. We already use it also for other entities.

src/services/TestDataService.ts
```diff
@@ -547,11 +511,7 @@ export class TestDataService {
      * @param salutationKey - The key of the salutation that should be used for the customer. Default is "mr".
      * @param salesChannel - The sales channel for which the customer should be registered.
      */
-    async createCustomer(
-        overrides: Partial<Customer> = {},
-        salutationKey = 'mr',
-        salesChannel: SalesChannel = this.defaultSalesChannel,
-    ): Promise<Customer> {
+    async createCustomer(overrides: Partial<Customer> = {}, salutationKey = 'mr', salesChannel: SalesChannel = this.defaultSalesChannel): Promise<Customer> {
```
yusufttur, line 506: Hi If line width set to 180, it should be in two lines because we have 200 char here. 🤔

src/services/TestDataService.ts
```diff
@@ -1417,16 +1392,20 @@ export class TestDataService {
...
+                    },
+                ],
             },
         });
         expect(response.ok()).toBeTruthy();
 
-        const { data: result } = (await response.json()) as { data: PaymentMethod[] };
+        const { data: result } = (await response.json()) as {
+            data: PaymentMethod[];
+        };
```
yusufttur: Why it is in more than one line? In total we have 90 chars.

yusufttur (changes requested):

src/services/TestDataService.ts
frobel: I guess because of code definition before, but I changed it back to one line and its fine now

yusufttur (approved):

PR #283, opened 2025-01-22
fix: add methods to suggest page and fix lineitems in homepage

ocavli:


vanpham-sw:
Please see the fail check test / test (shopware, http://localhost:8011) (pull_request) , 
I tried rerun it but still failed.

src/page-objects/storefront/SearchSuggest.ts
```diff
@@ -22,9 +23,22 @@ export class SearchSuggest extends Home implements PageObject {
...
         this.searchHeadline = page.locator ('.search-headline');
     }
 
+    async getTotalSearchResultCount(): Promise<number> {
+        const totalCountText = await this.searchResultTotal.textContent();
+        return parseInt(totalCountText?.trim().replace(/\D/g, '') || '0', 10);
+    }
+
+    async searchForTerm(searchTerm: string): Promise<void> {
+        await this.searchInput.fill(searchTerm);
```
vanpham-sw: I'm not sure we want to have this action here in the page objects class. According to the README.md: **They help you to organize page specific locators and provide helpers for interacting with a given page.**

I'll leave it up to you.

vanpham-sw (dismissed):

src/page-objects/storefront/SearchSuggest.ts
frobel: I'm also on the page of @vanpham-sw to keep the page object clean and stick on the purpose of a page object.

src/page-objects/storefront/Home.ts
```diff
@@ -32,7 +32,7 @@ export class Home implements PageObject {
         this.accountMenuButton = page.getByLabel('Your account');
         this.closeGuestSessionButton = page.locator('.account-aside-btn');
         this.productImages = page.locator('.product-image-wrapper');
-        this.productListItems = page.getByRole('listitem');
+        this.productListItems = page.locator('.product-box');
```
frobel, line 35: What is the reason for that change?

ocavli: With the old locator, it was selecting not only products but also other items as well. 
E.g. If we have 5 products on the listing, it selects 8 items with some others in the footer etc.

src/page-objects/storefront/SearchSuggest.ts
ocavli: should be fixed now 👍

vanpham-sw (approved):

src/page-objects/storefront/Home.ts
frobel: Ok I see your point.

frobel (approved):

PR #285, opened 2025-01-23
fix: generate a new access_token if existing one is expired

vanpham-sw:


vanpham-sw:
Hi @Phil23,
We have[ flaky test ](https://app.currents.dev/instance/gjvrCPjcw089QgXx)that looks like it's because the access token is expired, so I think we can update your isAuthenticated to check and regenerate the token for subsequence API call. Could you please help to take a look? Thanks! 
image

src/services/AdminApiContext.ts
```diff
@@ -123,31 +123,50 @@ export class AdminApiContext {
...
 
     isAuthenticated(): boolean {
-        // TODO: check token expiry
-        return !!this.options['access_token'];
+        if (!this.options['access_token']) {
+            return false;
+        }
+
+        const token = this.options['access_token'];
+        const decodedToken: { exp: number } = jwtDecode(token);
```
Phil23: We should move the decode to the writing of the token option and save the expiry date in a separate option, so we don't have to decode the token on every request.

Just out of my head without testing, but something like this:

```JavaScript
// authenticateWithClientCredentials method

const authData = (await authResponse.json()) as { access_token?: string };

const decodedToken: { exp: number } = jwtDecode(authData['access_token']);

contextOptions['access_token_exp'] = decodedToken.exp;
```

```JavaScript
// isAuthenticated method

if (!this.options['access_token'] || !this.options['access_token_exp']) {
    return false;
}

return this.options['access_token_exp'] > Math.floor(Date.now() / 1000);
```

src/services/AdminApiContext.ts
```diff
@@ -123,31 +123,50 @@ export class AdminApiContext {
...
+
+    async ensureAuthenticated(): Promise<void> {
+        if (!this.isAuthenticated()) {
+            this.options['access_token'] = await AdminApiContext.authenticateWithClientCredentials(this.context, this.options);
+            this.context = await AdminApiContext.createApiRequestContext(this.options);
+        }
     }
 
     async get<PAYLOAD>(url: string, options?: RequestOptions<PAYLOAD>): Promise<APIResponse> {
+        await this.ensureAuthenticated();
```
Phil23: Maybe another solution to consider: Instead of decoding the token locally and comparing the expiration date every time, we could just wait for the response here, and if the response is an unauthorized response code, we could fetch a new token and rerun the request. That is how the administration does it, if I am not mistaken.

Phil23 (changes requested):

frobel (approved):

pweyck:
Looks fine to me 👍 I'm just wondering: how long is the access token valid?

pweyck (approved):

vanpham-sw:
> Looks fine to me 👍 I'm just wondering: how long is the access token valid?

Hey, good question. cmiiw, I tried to find out by decoding an access token and it's 10 minutes. The test case ran for 3 seconds only so I don't know why the request returned:
```
{
  "errors": [
    {
      "code": "9",
      "status": "401",
      "title": "The resource owner or authorization server denied the request.",
      "detail": "Access token could not be verified"
    }
  ]
}
```
Any ideas?

pweyck:
> Hey, good question. cmiiw, I tried to find out by decoding an access token and it's 10 minutes

Did you use an integration or the normal admin login? I believe the lifetime differs and if it's just 10 minutes it should have caused issues already, because the test suite already takes longer, right? Or maybe all fixtures have been created before the token has expired?

> The test case ran for 3 seconds only so I don't know why the request returned:

I've no clue what causes this.

If you want to test your change you can change the ttl like [this](https://github.com/shopware/shopware/blob/25b69798549e8b0c6f63eeab3e40a3da4db19450/src/Core/Framework/Resources/config/packages/e2e/e2e.yaml#L31)

EDIT: you need to change that in the default config file, because the `e2e` app env is not used anymore (we should remove it), the ats uses `prod`. So that should be https://github.com/shopware/shopware/blob/25b69798549e8b0c6f63eeab3e40a3da4db19450/src/Core/Framework/Resources/config/packages/shopware.yaml

Phil23 (approved):
Looks good 👍

PR #286, opened 2025-01-24
Use string as scope

keulinho:
Using a list of scopes was not spec compliant: https://github.com/thephpleague/oauth2-server/issues/1469

With an update of our oauth lib this is no longer supported, leading to fails in the ATS: https://github.com/shopware/shopware/pull/5986

### DIFF
--- a/src/services/AdminApiContext.ts
+++ b/src/services/AdminApiContext.ts
@@ -89,7 +89,7 @@ export class AdminApiContext {
                 grant_type: 'client_credentials',
                 client_id: options.client_id,
                 client_secret: options.client_secret,
-                scope: ['write'],
+                scope: 'write',
             },
         });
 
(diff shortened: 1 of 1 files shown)


src/services/AdminApiContext.ts
```diff
@@ -89,7 +89,7 @@ export class AdminApiContext {
                 grant_type: 'client_credentials',
                 client_id: options.client_id,
                 client_secret: options.client_secret,
-                scope: ['write'],
+                scope: 'write',
```
pweyck, line 92: we need to support older shopware versions (at least the LTS), so we should try both variations, if we can't detect the version (I believe that is only possible after auth)

keulinho: I know, it should have always worked to provide plain string therefore there should be no need to switch the behaviour

PR #298, opened 2025-02-03
feat: add abstractions for flow builder create test

### LINKED ISSUE #8823: ATS: Flow Builder Create test
### User story

As an admin user, I want to create new flows, so that I can automate certain business workflows

```
GIVEN the admin user is logged in
WHEN the user creates a new flow in the Flow Builder
AND defines a trigger
AND adds a condition
AND adds actions for both branches of the condition
THEN the user can save the flow correctly
AND validate that the defined flow is structured correctly
```

### Acceptance criteria

Test running

GitEvil:
Changes to ATS for NEXT-39672 - Flow Builder Create test


src/page-objects/administration/FlowBuilderListing.ts
```diff
@@ -42,7 +44,13 @@ export class FlowBuilderListing implements PageObject {
     async getLineItemByFlowName(flowName: string): Promise<Record<string, Locator>> {
         const lineItem = this.page.locator('.sw-data-grid__row').filter({ hasText: flowName });
         const flowSelectionCheckbox = lineItem.locator('.sw-data-grid__cell--selection').locator('.sw-field__checkbox');
-        const flowActiveCheckmark = lineItem.locator('.sw-data-grid__cell--active').locator('.icon--regular-checkmark-xs');
+        let flowActiveCheckmark
+        if (satisfies(this.instanceMeta.version, '<6.6.9999999')) {
```
sguder, line 48: Should be 6.7

src/page-objects/administration/FlowBuilderDetail.ts
```diff
@@ -1,19 +1,25 @@
...
+import { HelperFixtureTypes } from '../../fixtures/HelperFixtures';
 
-export class FlowBuilderDetail implements PageObject {
+export class FlowBuilderDetail extends FlowBuilderCreate implements PageObject {
+
+    public readonly successMessage: Locator;
+
+    constructor(public readonly page: Page, public readonly instanceMeta: HelperFixtureTypes['InstanceMeta']) {
+        super(page, instanceMeta);
+        if (satisfies(instanceMeta.version, '<6.6.9999999')) {
```
sguder: Should be 6.7

src/page-objects/administration/FlowBuilderListing.ts
sguder: 6.7 too

src/tasks/shop-admin/Flow/CreateFlow.ts
```diff
@@ -0,0 +1,59 @@
...
+                await ShopAdmin.expects(AdminFlowBuilderCreate.page.locator('.sw-select-result-list__item-list')).toBeVisible();
+                await AdminFlowBuilderCreate.page.locator('.sw-select-result-list__content').getByRole('listitem').filter({ hasText: `${flowConfig.falseAction}` }).click();
+                await ShopAdmin.expects(AdminFlowBuilderCreate.page.locator('.sw-modal__title')).toHaveText('Add tag');
+                await AdminFlowBuilderCreate.page.locator('.sw-select__selection').getByLabel('Tags').click();
+                await ShopAdmin.expects(AdminFlowBuilderCreate.page.locator('.sw-select-result-list__item-list')).toBeVisible();
+                await AdminFlowBuilderCreate.page.locator('.sw-select-result-list__content').getByRole('listitem').filter({ hasText: `${flowConfig.falseActionIdentifier}` }).click();
+                await AdminFlowBuilderCreate.modalAddButton.click();
+                await ShopAdmin.expects(AdminFlowBuilderCreate.page.locator('.sw-flow-sequence__false-block').locator('.sw-flow-sequence-action__actions')).toContainText(`Tag: ${flowConfig.falseActionIdentifier}`);
```
sguder: Why all these locators here instead on a page object?

src/tasks/shop-admin/Flow/CheckFlow.ts
```diff
@@ -0,0 +1,39 @@
...
+                // Make sure there is only one condition
+                await ShopAdmin.expects(AdminFlowBuilderDetail.page.locator('.sw-flow-sequence-condition__rule-header')).toHaveText(`${flowConfig.condition}`);
+                // Make sure there are no delays
+                await ShopAdmin.expects(AdminFlowBuilderDetail.page.locator('.sw-flow-delay-action__delay_card')).not.toBeVisible();
+                // Make sure there is only one section
+                await ShopAdmin.expects(AdminFlowBuilderDetail.page.locator('.sw-flow-detail-flow__position-connection')).toBeVisible();
+                // Make sure there are only desired actions present after the condition
+                await ShopAdmin.expects(AdminFlowBuilderDetail.page.locator('.sw-flow-sequence__true-block').locator('.sw-flow-sequence-action__action-description')).toContainText(`${flowConfig.trueActionIdentifier}`);
+                await ShopAdmin.expects(AdminFlowBuilderDetail.page.locator('.sw-flow-sequence__false-block').locator('.sw-flow-sequence-action__action-description')).toContainText(`${flowConfig.falseActionIdentifier}`);
```
sguder: Why all these locators here instead on a page object?

sguder (changes requested):

src/page-objects/administration/FlowBuilderCreate.ts
```diff
@@ -1,14 +1,64 @@
...
+        if (satisfies(instanceMeta.version, '<6.6.9999999')) {
+            this.triggerSelectField = page.locator('.sw-flow-detail-flow__trigger-card').getByRole('textbox');
+        } else {
+            this.triggerSelectField = page.locator('.sw-flow-detail-flow__trigger-card').getByRole('textbox');
+        }
+        this.addActionField = page.locator('.sw-flow-sequence-action__content').locator('.sw-single-select__selection');
+        if (satisfies(instanceMeta.version, '<6.6.9999999')) {
+            this.modalAddButton = page.locator('.sw-button--primary').getByText('Add action');
+        } else {
+            this.modalAddButton = page.locator('.mt-button--primary').getByText('Add action');
```
frobel: Would be more organised when you put all if related stuff together and not using an IF per Locator

src/page-objects/administration/FlowBuilderListing.ts
```diff
@@ -42,7 +44,13 @@ export class FlowBuilderListing implements PageObject {
...
+            flowActiveCheckmark = lineItem.locator('.sw-data-grid__cell--active').locator('.icon--regular-checkmark-xs');
+        } else {
+            flowActiveCheckmark = lineItem.locator('.sw-data-grid__cell--active').getByTestId('mt-icon__regular-checkmark-xs');
+        }
+        //const flowActiveCheckmark = lineItem.locator('.sw-data-grid__cell--active').locator('.icon--regular-checkmark-xs');
```
frobel: comment can be deleted

src/tasks/shop-admin/Flow/CheckFlow.ts
```diff
@@ -0,0 +1,39 @@
+import { test as base } from '@playwright/test';
+import type { Task } from '../../../types/Task';
+import type { FixtureTypes } from '../../../types/FixtureTypes';
+import { FlowConfig } from '../../../types/ShopwareTypes';
+
+export const CheckFlow = base.extend<{ CheckFlow: Task }, FixtureTypes>({
```
frobel: Does it make more sense to make the validation in the Test and noch in the Task?

frobel (changes requested):

src/tasks/shop-admin/Flow/CheckFlow.ts
GitEvil: Totally!

src/services/ShopwareUIHelpers.ts
```diff
@@ -0,0 +1,38 @@
+import { type Page, type Locator } from '@playwright/test';
+
+/**
+ * Returns the locator for a list item on a select field's result list.
```
frobel: Somehow I don't like the approach of introducing a new silo for lots or functions where the purpose of the class is not specific enough. Let's think about it again. 

I have already an idea:
- create a service like CustomFieldCard which inject a page object so that we can manage there some card specific functions like getCustomFieldCardContent(). And use the functions within page objects which want to use it with:

```
import { getCustomFieldCardContent } from '../../services/ShopwareUIHelpers.ts';

export class ProductDetail implements PageObject {
    // ...

    public getCustomFieldCardContent(selectField: Locator, listItem: string) {
        return getCustomFieldCardContent(this.page, locator, listItem);
    }
}
```

so within your test you could call then:

`const customFieldLocators = await AdminProductDetail.getCustomFieldCardContent(selectField, listItem: string);`

@vanpham-sw and @yusufttur what do you think about it?

frobel (changes requested):

src/services/ShopwareUIHelpers.ts
GitEvil: @frobel I purposefully added the function to the UIHelpers class so that it would not have to be declared in every page object. With this test alone, we check 3-5 different detail pages and would thus have to add the function to each page object. I just found it easier to put it into the Helpers class once.

vanpham-sw: @GitEvil Even though this is a clever solution, I'd agree with Florian not to do it this way. This can lead to complexity if the functions are not cohesively related or if the class becomes a "catch-all" for unrelated functionalities. If the purpose is not specific, the class may become overly complex and difficult to maintain.

Have you tried below approach?
- Create a CustomFieldCard page object class contains getResultListItem & getCustomFieldCardContent
- detail pages: class ProductDetail extends CustomFieldCard implements PageObject

frobel: @vanpham-sw for me the `extends` definition is something different. Extends means derived class or subclass. So for me it is just a kind of module which we are using within a page object. So to make it more clear for the developer I would say, that as a devolper it is possible to include modules to page objects via import. Like I say above:

```
// Product Detail.ts
import { getCustomFieldCardContent } from '../../services/modules/customFieldCard.ts';
public getCustomFieldCardContent(selectField: Locator, listItem: string) {
        return getCustomFieldCardContent(this.page, locator, listItem);
    }
```

and in the customFieldCard.ts looks like the following:

```
//  customFieldCard.ts within services/modules
/**
 * Returns locators for assigned custom fields on entity detail pages.
 *
 * @param pageObject
 * @param customFieldSetName
 * @param customFieldName
 * @returns Custom field locators on entity detail pages.
 */
export const getCustomFieldCardContent = async (pageObject: Page, customFieldSetName: string, customFieldName: string): Promise<Record<string, Locator>> => {

    const customFieldCard = pageObject.locator('.sw-card').filter({ hasText: 'Custom fields' });
    const customFieldSetTab = customFieldCard.getByText(customFieldSetName);
    const customFieldLabel = customFieldCard.locator('.sw-custom-field-set-renderer').locator('.sw-field__label').getByText(customFieldName);
    const customFieldSelect = customFieldCard.locator(`.sw-custom-field-set-renderer-tab-content__${customFieldSetName}`);

    return {
        customFieldCard: customFieldCard,
        customFieldSetsTab: customFieldSetTab,
        customFieldLabel: customFieldLabel,
        customFieldSelect: customFieldSelect,
    }
```

frobel: We can also discuss if the place is the right one under `services/modules`  or better `modules` as a new folder :thinking:

vanpham-sw: Yes, you made good point here, agree that this is more like an module where we import than an **extends**. I believe with this approach we can import multiple modules, which is also future-proof 👍

vanpham-sw: > We can also discuss if the place is the right one under `services/modules` or better `modules` as a new folder 🤔

why not `page-objects/modules` ?

yusufttur: We can gather all UI helpers under the services/UIHelpers directory. I like having separate files for each helper and importing them into page objects.

Phil23: Ui interaction is definitely something that should go into a page object. One approach you could also consider is to change the `PageObject` interface into an abstract class that shares some main functionalities the other page object instances inherit or adjust to their use case. You could for example have two different abstract classes for Admin and Storefront, from which the specific page objects inherit.

```TypeScript
// Abstract class

class AdminPageObject {

    public getCustomFieldCardContent(selectField: Locator, listItem: string) {
        ....
    }
}
```

```TypeScript
// Specific Page Object

class ProductDetail extends AdminPageObject {

    constructor() {
        super();
    }
}
```

You could do a lot of interesting things with this by combining specific page logic with the super class. In addition we would have a better identification of page object domains like:

```TypeScript 
ProductDetail instanceof AdminPageObject
```

frobel: Thanks, @Phil23, for your input. We will take your suggestion into account for other situations. As a team, we decided to go with the module approach, so we will implement module classes that contain locators that could be used from different page objects implemented via a get method.

src/page-objects/administration/FlowBuilderListing.ts
```diff
@@ -42,12 +44,16 @@ export class FlowBuilderListing implements PageObject {
...
         const lineItem = this.page.locator('.sw-data-grid__row').filter({ hasText: flowName });
         const flowSelectionCheckbox = lineItem.locator('.sw-data-grid__cell--selection').locator('.sw-field__checkbox');
-        const flowActiveCheckmark = lineItem.locator('.sw-data-grid__cell--active').locator('.icon--regular-checkmark-xs');
+        let flowActiveCheckmark
+        if (satisfies(this.instanceMeta.version, '<6.7')) {
+            flowActiveCheckmark = lineItem.locator('.sw-data-grid__cell--active').locator('.icon--regular-checkmark-xs');
+        } else {
+            flowActiveCheckmark = lineItem.locator('.sw-data-grid__cell--active').getByTestId('sw-icon__regular-checkmark-xs');
+        }
         const flowDisabledCheckmark = lineItem.locator('.sw-data-grid__cell--active').locator('.icon--regular-times-s');
```
frobel: Why does that locator not differ between 6.6 and 6.7? like `.icon--regular-checkmark-xs` :thinking: 

I would also say there is a different in 6.6. and 6.7. (not tested)

vanpham-sw (approved):

GitEvil:
Issue: https://github.com/shopware/shopware/issues/8823

src/tasks/shop-admin/Flow/CreateFlow.ts
```diff
@@ -23,20 +23,46 @@ export const CreateFlow = base.extend<{ CreateFlow: Task }, FixtureTypes>({
...
+                //await (await AdminFlowBuilderCreate.getSelectFieldListitem(AdminFlowBuilderCreate.trueBlockActionSelectField, `${flowConfig.trueAction}`)).click();
                 await ShopAdmin.expects(AdminFlowBuilderCreate.mailSendModal).toBeVisible();
-                await (await AdminFlowBuilderCreate.getSelectFieldListitem(AdminFlowBuilderCreate.mailSendModalTemplateSelectField, `${flowConfig.trueActionIdentifier}`)).click();
+                await AdminFlowBuilderCreate.mailSendModalTemplateSelectField.click();
+                await AdminFlowBuilderCreate.page.locator('.sw-select-result-list__item-list').waitFor({ state: 'visible' });
+                await AdminFlowBuilderCreate.page.locator('.sw-select-result-list__content').getByRole('listitem').filter({ hasText: `${flowConfig.trueActionIdentifier}`}).click();
+                //await (await AdminFlowBuilderCreate.getSelectFieldListitem(AdminFlowBuilderCreate.mailSendModalTemplateSelectField, `${flowConfig.trueActionIdentifier}`)).click();
+                // Hide symfony toolbar to make modalAddButton accessible
+                await AdminFlowBuilderCreate.page.addStyleTag({
```
yusufttur, line 42: Why do we need to add this here again? I think this should apply to all pages?
https://github.com/shopware/acceptance-test-suite/blob/d5ad962a0daf2ceabc93436072f72a894e8b78d7/src/fixtures/PageContexts.ts#L50

src/tasks/shop-admin/Flow/CreateFlow.ts
```diff
@@ -23,20 +23,46 @@ export const CreateFlow = base.extend<{ CreateFlow: Task }, FixtureTypes>({
...
+                    content: `
+                        .sf-toolbar {
```
yusufttur: Why do we need to add this here again? I think this should apply to all pages?
https://github.com/shopware/acceptance-test-suite/blob/d5ad962a0daf2ceabc93436072f72a894e8b78d7/src/fixtures/PageContexts.ts#L50

GitEvil: That is where I found the code. It did not remove the toolbar in my local environment, so I put it here again.

frobel: It seems to be just a local env problem, if not that out pipelines should also brake or?

GitEvil: Okay, this was implemented 10 months ago. It should not be a problem, but I need to check why it is still there in my environments next week. 🤔

src/page-objects/administration/FlowBuilderListing.ts
GitEvil: Good catch, the first locator works fine for 6.6 and 6.7 respectively, so I removed the instanceMeta.

src/tasks/shop-admin/Flow/CreateFlow.ts
GitEvil: Since this is just a problem I have locally, I removed the code.

src/tasks/shop-admin/Flow/CreateFlow.ts
```diff
@@ -0,0 +1,63 @@
+import { test as base } from '@playwright/test';
+import type { Task } from '../../../types/Task';
+import type { FixtureTypes } from '../../../types/FixtureTypes';
+import { FlowConfig } from '../../../types/ShopwareTypes';
+
+export const CreateFlow = base.extend<{ CreateFlow: Task }, FixtureTypes>({
```
vanpham-sw, line 6: Just wonder if the created flow is cleaned up (ideally by addCreatedRecord)

src/tasks/shop-admin/Flow/CreateFlow.ts
```diff
@@ -0,0 +1,63 @@
...
+                await AdminFlowBuilderCreate.flowTab.click();
+                // Select trigger
+                await AdminFlowBuilderCreate.triggerSelectField.fill(flowConfig.triggerSearchTerm);
+                await AdminFlowBuilderCreate.triggerSelectField.press('Enter');
+                // Add condition
+                await AdminFlowBuilderCreate.sequenceSelectorConditionButton.click();
+                await AdminFlowBuilderCreate.conditionSelectField.click();
+                await ShopAdmin.expects(AdminFlowBuilderCreate.page.locator('.sw-select-result-list__item-list')).toBeVisible();
+                await AdminFlowBuilderCreate.page.locator('.sw-select-result-list__content').getByRole('listitem').filter({ hasText: `${flowConfig.condition}`}).click();
+                //await (await AdminFlowBuilderCreate.getSelectFieldListitem(AdminFlowBuilderCreate.conditionSelectField, `${flowConfig.condition}`)).click();
```
vanpham-sw, line 30: Please remove all unnecessary commented code

src/page-objects/administration/FlowBuilderCreate.ts
```diff
@@ -1,32 +1,103 @@
...
+        this.mailSendModalTemplateSelectField = page.locator('.sw-flow-mail-send-modal').locator('.sw-entity-single-select__selection');
+        this.falseBlockAddConditionButton = page.locator('.sw-flow-sequence__false-block').getByText('Add condition (IF)');
+        this.falseBlockAddActionButton = page.locator('.sw-flow-sequence__false-block').getByText('Add action (THEN)');
+        this.falseBlockActionSelectField = page.locator('.sw-flow-sequence__false-block').locator('.sw-single-select');
+        this.falseBlockActionDescription = page.locator('.sw-flow-sequence__false-block').locator('.sw-flow-sequence-action__actions');
+        this.tagModal = page.locator('.sw-flow-tag-modal');
+        this.tagModalTagsSelectField = page.locator('.sw-flow-tag-modal').locator('.sw-select__selection').getByLabel('Tags');
+        this.delayCard = page.locator('.sw-flow-delay-action__delay_card');
+        this.sequenceSeparator = page.locator('.sw-flow-detail-flow__position-connection');
+        this.newFlowHeader = page.getByRole('heading', { name: 'New flow' })
```
frobel:
```suggestion
        this.newFlowHeader = page.getByRole('heading', { name: 'New flow' });
```

src/page-objects/administration/FlowBuilderCreate.ts
```diff
@@ -1,32 +1,103 @@
...
     }
 
     url(flowId?: string, tabName = 'general') {
         if (!flowId || flowId === '') {
             return `#/sw/flow/create/${tabName}`;
         }
         return `#/sw/flow/create/${flowId}/${tabName}`;
+
+    }
+    async getSelectedTrigger(): Promise<string> {
```
frobel: Is that possibly a "module" which we can also use in other pages like product detail tooltip etc?

src/tasks/shop-admin/Flow/CreateFlow.ts
frobel: good point

src/page-objects/administration/FlowBuilderCreate.ts
```diff
@@ -1,32 +1,103 @@
...
+        await this.page.locator('.sw-flow-detail-flow__trigger-card').getByRole('textbox').hover();
+        const tooltip = await this.page.waitForSelector('.sw-tooltip');
```
sguder: You can disable the lint warning with a comment, if you need the waitForSelector

src/tasks/shop-admin/Flow/CreateFlow.ts
sguder: I think it won't, because it isn't creatd by the TestDataService

src/tasks/shop-admin/Flow/CreateFlow.ts
```diff
@@ -0,0 +1,63 @@
...
+                await ShopAdmin.expects(AdminFlowBuilderCreate.page.locator('.sw-select-result-list__item-list')).toBeVisible();
+                await AdminFlowBuilderCreate.page.locator('.sw-select-result-list__content').getByRole('listitem').filter({ hasText: `${flowConfig.falseActionIdentifier}`}).click();
+                //await (await AdminFlowBuilderCreate.getSelectFieldListitem(AdminFlowBuilderCreate.tagModalTagsSelectField, `${flowConfig.falseActionIdentifier}`)).click();
+                await AdminFlowBuilderCreate.modalAddButton.click();
+                await ShopAdmin.expects(AdminFlowBuilderCreate.falseBlockActionDescription).toContainText(`Tag: ${flowConfig.falseActionIdentifier}`);
+                await AdminFlowBuilderCreate.saveButton.click();
+                await ShopAdmin.expects(AdminFlowBuilderDetail.successMessage).toBeVisible();
+            }
+        };
+        await use(task);
```
sguder, line 69: As mentioned above you should insert code to delete the created flow after use

sguder (changes requested):

src/tasks/shop-admin/Flow/CreateFlow.ts
```diff
@@ -0,0 +1,63 @@
...
+                // Switch to flow tab
+                await AdminFlowBuilderCreate.flowTab.click();
+                // Select trigger
+                await AdminFlowBuilderCreate.triggerSelectField.fill(flowConfig.triggerSearchTerm);
+                await AdminFlowBuilderCreate.triggerSelectField.press('Enter');
+                // Add condition
+                await AdminFlowBuilderCreate.sequenceSelectorConditionButton.click();
+                await AdminFlowBuilderCreate.conditionSelectField.click();
+                await ShopAdmin.expects(AdminFlowBuilderCreate.page.locator('.sw-select-result-list__item-list')).toBeVisible();
+                await AdminFlowBuilderCreate.page.locator('.sw-select-result-list__content').getByRole('listitem').filter({ hasText: `${flowConfig.condition}`}).click();
```
frobel: You have some reusable code here for example:
//condition
await AdminFlowBuilderCreate.page.locator('.sw-select-result-list__content').getByRole('listitem').filter({ hasText: `${flowConfig.condition}`}).click();

//trueAction
await AdminFlowBuilderCreate.page.locator('.sw-select-result-list__content').getByRole('listitem').filter({ hasText: `${flowConfig.trueAction}` }).click();

// trueActionIdentifier
await AdminFlowBuilderCreate.page.locator('.sw-select-result-list__content').getByRole('listitem').filter({ hasText: `${flowConfig.trueActionIdentifier}`}).click();

etc. 

So you could at least create on locator so that you can write:
await AdminFlowBuilderCreate.**resultListItem**.filter(...).click()

src/tasks/shop-admin/Flow/CreateFlow.ts
```diff
@@ -0,0 +1,63 @@
...
+                await AdminFlowBuilderCreate.page.locator('.sw-select-result-list__item-list').waitFor({ state: 'visible' });
+                await AdminFlowBuilderCreate.page.locator('.sw-select-result-list__content').getByRole('listitem').filter({ hasText: `${flowConfig.trueActionIdentifier}`}).click();
+                //await (await AdminFlowBuilderCreate.getSelectFieldListitem(AdminFlowBuilderCreate.mailSendModalTemplateSelectField, `${flowConfig.trueActionIdentifier}`)).click();
+                await AdminFlowBuilderCreate.modalAddButton.click();
+                await ShopAdmin.expects(AdminFlowBuilderCreate.trueBlockActionDescription).toContainText(`${flowConfig.trueActionIdentifier}`);
+                // Add action to condition false block
+                await AdminFlowBuilderCreate.falseBlockAddActionButton.click();
+                await ShopAdmin.expects(AdminFlowBuilderCreate.falseBlockActionSelectField).toBeVisible();
+                await AdminFlowBuilderCreate.falseBlockActionSelectField.click();
+                await ShopAdmin.expects(AdminFlowBuilderCreate.page.locator('.sw-select-result-list__item-list')).toBeVisible();
```
frobel: same here. You could create one locator abstraction for that:

await AdminFlowBuilderCreate.page.locator('.sw-select-result-list__content').getByRole('listitem')
Line: 47 and Line 52

frobel (changes requested):

src/page-objects/administration/FlowBuilderCreate.ts
GitEvil: This is rather a workaround to get the trigger name, which consists of separate parts. I don't know where on the product detail page we would want to use this. Can you explain where you think this would make sense to use?

src/page-objects/administration/FlowBuilderCreate.ts
GitEvil: ✅

src/page-objects/administration/FlowBuilderCreate.ts
GitEvil: Found a way around using this, thanks. ✅

src/tasks/shop-admin/Flow/CreateFlow.ts
GitEvil: Flow is being deleted now. Thanks for pointing that out. 👍

src/tasks/shop-admin/Flow/CreateFlow.ts
GitEvil: Originally, I wanted to use the getSelectFieldListitem method here. But it doesn't work until the select fields are migrated to their new Meteor version. That won't happen until the end of the year. But you are right, I am going to create a locator.

src/tasks/shop-admin/Flow/CreateFlow.ts
GitEvil: All I could do was abstract the `page.locator('.sw-select-result-list__content')`. If I include the getByRole('listitem'), the test fails because it finds multiple list items.

src/tasks/shop-admin/Flow/CreateFlow.ts
GitEvil: Done. ✅

src/tasks/shop-admin/Flow/CreateFlow.ts
GitEvil: @frobel @sguder The commented lines can replace the respective three lines above as soon as the select field component is moved to Meteor. Is it worth keeping it here? In my opinion, this makes it easier to apply this improvement in the future.

src/page-objects/administration/FlowBuilderCreate.ts
GitEvil: Created the module anyway. :D

src/tasks/shop-admin/Flow/CreateFlow.ts
sguder: You should comment it with a TODO then, to make it clear

src/tasks/shop-admin/Flow/CreateFlow.ts
sguder: Apart from that I would strongly suggest keeping locators only in the respective page object to avoid having to scoop through many files if that locator has to be changed one day. That's why we have page objects

sguder (approved):

src/page-objects/administration/FlowBuilderCreate.ts
```diff
@@ -1,32 +1,106 @@
...
     }
 
     url(flowId?: string, tabName = 'general') {
         if (!flowId || flowId === '') {
             return `#/sw/flow/create/${tabName}`;
         }
         return `#/sw/flow/create/${flowId}/${tabName}`;
+
+    }
+    async getSelectedTrigger(): Promise<string> {
```
frobel: Could you use here the getTooltipText() from './modules/TooltipText';

src/page-objects/administration/FlowBuilderCreate.ts
```diff
@@ -1,32 +1,106 @@
...
+        this.activeSwitch = page.locator('.sw-flow-detail-general__general-active').getByLabel('Active');
+        this.sequenceSelectorConditionButton = page.locator('.sw-flow-sequence-selector__actions').getByText('Add condition (IF)');
+        this.conditionSelectField = page.locator('.sw-flow-sequence-condition__select');
+        this.conditionRule = page.locator('.sw-flow-sequence-condition__rule-header');
+        this.sequenceSelectorActionButton = page.locator('.sw-flow-sequence-selector__actions').getByText('Add action (THEN)');
+        this.actionSelectField = page.locator('.sw-flow-sequence-action__content').locator('.sw-single-select__selection');
+        this.selectFieldResultList = page.locator('.sw-select-result-list__item-list');
+        this.trueBlockAddConditionButton = page.locator('.sw-flow-sequence__true-block').getByText('Add condition (IF)');
+        this.trueBlockAddActionButton = page.locator('.sw-flow-sequence__true-block').getByText('Add action (THEN)');
+        this.trueBlockActionSelectField = page.locator('.sw-flow-sequence__true-block').locator('.sw-single-select');
```
frobel: you can abstract/summarize all `.sw-flow-sequence__true-block` to one locator and just use that locator variable in other locators like:

const flowSequenceTrueBlock = page.locator('.sw-flow-sequence__true-block');
this.trueBlockAddActionButton = flowSequenceTrueBlock.getByText('Add action (THEN)');

src/page-objects/administration/FlowBuilderCreate.ts
```diff
@@ -1,32 +1,106 @@
...
+        this.triggerSelectField = page.locator('.sw-flow-detail-flow__trigger-card').getByRole('textbox');
+        this.flowTab = page.locator('.sw-tabs__content').locator('.sw-flow-detail__tab-flow');
+        if (satisfies(instanceMeta.version, '<6.7')) {
+            this.modalAddButton = page.locator('.sw-button--primary').getByText('Add action');
+        } else {
+            this.modalAddButton = page.locator('.mt-button--primary').getByText('Add action');
+        }
+        this.nameField = page.locator('.sw-flow-detail-general__general-name').getByLabel('Name');
+        this.descriptionField = page.locator('.sw-flow-detail-general__general-description').getByLabel('Description');
+        this.priorityField = page.locator('.sw-flow-detail-general__general-priority').getByLabel('Priority');
```
frobel: mhh is that locator definition really necessary. Is it not enough to just use page.getByLabel('Priority') ? Why you created it in that way :thinking:

frobel (changes requested):

src/page-objects/administration/FlowBuilderCreate.ts
GitEvil: It actually is not necessary. No idea why I went that way, probably was just glad that it worked. :D Removing unnecessary locators.

src/page-objects/administration/FlowBuilderCreate.ts
GitEvil: 100%! Already using it, but I think I forgot to remove/rewrite it here. 🙈 Great catch 👍

src/page-objects/administration/FlowBuilderDetail.ts
```diff
@@ -13,25 +14,34 @@ export class FlowBuilderDetail extends FlowBuilderCreate implements PageObject {
...
         }
         this.saveButton = page.locator('.sw-flow-detail__save');
         this.flowTab = page.locator('.sw-flow-detail__tab-flow');
         this.templateName = page.getByLabel('Name');
+        this.actionContentTag = page.locator('.sw-flow-sequence-action__content').locator('.tag');
     }
 
-    url(flowId: string, tabName = 'general') {
+    url(flowId?: string, tabName = 'general') {
         return `#/sw/flow/detail/${flowId}/${tabName}`
```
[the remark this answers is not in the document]
GitEvil: Did not follow suggestion, copied from FlowBuilderCreate instead.

src/page-objects/administration/FlowBuilderCreate.ts
```diff
@@ -1,26 +1,95 @@
-import type { Page, Locator } from '@playwright/test';
+import type { Locator, Page } from '@playwright/test';
 import type { PageObject } from '../../types/PageObject';
+import { satisfies } from 'compare-versions';
+import { HelperFixtureTypes } from '../../fixtures/HelperFixtures';
+import { getSelectFieldListitem } from './modules/SelectFieldListitem';
+import { getTooltipText } from './modules/TooltipText';
```
frobel: 'getTooltipText' is defined but never used

src/page-objects/administration/FlowBuilderCreate.ts
```diff
@@ -1,26 +1,95 @@
...
+            this.modalAddButton = page.locator('.mt-button--primary').getByText('Add action');
+        }
+        this.nameField = page.getByLabel('Name');
+        this.descriptionField = page.getByLabel('Description');
+        this.priorityField = page.getByLabel('Priority');
+        this.activeSwitch = page.getByLabel('Active');
+        this.sequenceSelectorConditionButton = page.locator('.sw-flow-sequence-selector__actions').getByText('Add condition (IF)');
+        this.conditionSelectField = page.locator('.sw-flow-sequence-condition__select');
+        this.conditionRule = page.locator('.sw-flow-sequence-condition__rule-header');
+        this.sequenceSelectorActionButton = page.locator('.sw-flow-sequence-selector__actions').getByText('Add action (THEN)');
```
frobel: here it is not enough to just take `page.getByText(Add action (Then))`

src/page-objects/administration/FlowBuilderCreate.ts
```diff
@@ -1,26 +1,95 @@
...
+        if (satisfies(instanceMeta.version, '<6.7')) {
+            this.modalAddButton = page.locator('.sw-button--primary').getByText('Add action');
+        } else {
+            this.modalAddButton = page.locator('.mt-button--primary').getByText('Add action');
+        }
+        this.nameField = page.getByLabel('Name');
+        this.descriptionField = page.getByLabel('Description');
+        this.priorityField = page.getByLabel('Priority');
+        this.activeSwitch = page.getByLabel('Active');
+        this.sequenceSelectorConditionButton = page.locator('.sw-flow-sequence-selector__actions').getByText('Add condition (IF)');
```
frobel: here it is not enough to just take `page.getByText(Add condition (If))`

frobel (changes requested):

frobel (approved):

PR #299, opened 2025-02-03
feat: add functions for changing the payment method on an existing order

sguder:


src/services/TestDataService.ts
```diff
@@ -100,7 +100,7 @@ export class TestDataService {
      *
      * @private
      */
-    private highPriorityEntities = ['order', 'product', 'landing_page', 'shipping_method', 'sales_channel_domain', 'sales_channel_currency', 'sales_channel_country', 'customer'];
+    private highPriorityEntities = ['order', 'product', 'landing_page', 'shipping_method', 'sales_channel_domain', 'sales_channel_currency', 'sales_channel_country', 'sales_channel_payment_method', 'customer'];
```
yusufttur, line 103: Just to know, how do we decide if an entity is high prio?

src/services/TestDataService.ts
```diff
@@ -1212,6 +1212,39 @@ export class TestDataService {
...
+                        },
+                    ],
+                },
+            },
+        });
+        expect(syncSalesChannelResponse.ok()).toBeTruthy();
+
+        const { data: salesChannel } = await syncSalesChannelResponse.json();
+
+        this.addCreatedRecord('sales_channel_currency', {
```
yusufttur: Why do you add delete it from sales_channel_currency? I guess it should be sales_channel_payment_method.

yusufttur (changes requested):

src/services/TestDataService.ts
frobel: If you figure out the the deletion is not possible of that entity because it is used from another entity

yusufttur: Ah okay then they got deleted at last.

sguder: No, they get deleted first

src/services/TestDataService.ts
sguder: Ah of course

yusufttur (approved):

ravenokko (approved):

src/services/TestDataService.ts
ravenokko, line 1: the `getBasicPaymentMethodStruct` by default does not seem to include the `Allow payment change after checkout` option. We should first validate it is active for at least the first payment method in your test as part of your test data configuration.


Screenshot 2025-02-03 at 17 28 37

ravenokko (changes requested):

src/services/TestDataService.ts
sguder: That's why I set it in the test:
TestDataService.createBasicPaymentMethod({ afterOrderEnabled: true });

I think in a general function that just has to return the element of a payment method on a page it is not the job of that function to ensure that this is set.

sguder: Oh and the first payment method is a standard pre-installed payment method. Also this option is only relevant for switching to the payment method, so for the initial payment method it's totally irrelevant

ravenokko (approved):

PR #300, opened 2025-02-04
feat: add new address

vanpham-sw:
https://shopware.atlassian.net/browse/NEXT-40295
Platform PR: https://github.com/shopware/shopware/pull/6589

### DIFF
--- a/src/page-objects/storefront/AccountAddresses.ts
+++ b/src/page-objects/storefront/AccountAddresses.ts
@@ -7,6 +7,8 @@ export class AccountAddresses implements PageObject {
     public readonly editShippingAddressButton: Locator;
     public readonly useDefaultBillingAddressButton: Locator;
     public readonly useDefaultShippingAddressButton: Locator;
+    public readonly deliveryNotPossibleAlert: Locator;
+    public readonly otherShippingAddress: Locator;
 
     constructor(public readonly page: Page) {
         this.addNewAddressButton = page.getByRole('link', { name: /Add (new )?address/ });
@@ -14,6 +16,9 @@ export class AccountAddresses implements PageObject {
         this.editShippingAddressButton = page.getByRole('link', { name: 'Edit address' }).nth(1);
         this.useDefaultBillingAddressButton = page.getByRole('button', { name: 'Use as default billing address' });
         this.useDefaultShippingAddressButton = page.getByRole('button', { name: 'Use as default shipping address' });
+        const otherAddress = page.locator('.other-address');
+        this.deliveryNotPossibleAlert = otherAddress.getByText('A delivery to this country is not possible.');
+        this.otherShippingAddress = otherAddress.locator('.address').nth(1);
     }
 
     url() {
--- a/src/page-objects/storefront/AccountAddresses.ts
+++ b/src/page-objects/storefront/AccountAddresses.ts
@@ -1,5 +1,7 @@
 import type { Page, Locator } from '@playwright/test';
 import type { PageObject } from '../../types/PageObject';
+import { HelperFixtureTypes } from '../../fixtures/HelperFixtures';
+import { satisfies } from 'compare-versions';
 
 export class AccountAddresses implements PageObject {
     public readonly addNewAddressButton: Locator;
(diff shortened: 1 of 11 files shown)


src/services/TestDataService.ts
```diff
@@ -100,7 +100,7 @@ export class TestDataService {
      *
      * @private
      */
-    private highPriorityEntities = ['order', 'product', 'landing_page', 'shipping_method', 'sales_channel_domain', 'sales_channel_currency', 'sales_channel_country', 'customer'];
+    private highPriorityEntities = ['order', 'product', 'landing_page', 'shipping_method', 'sales_channel_domain', 'sales_channel_currency', 'sales_channel_country', 'customer', 'customer_address'];
```
vanpham-sw: I'm having an issue cleaning up country data in after hook(409 conflict error), despite having added **customer_address** here.

yusufttur: Somehow the address data is still in use when you try to delete it.

src/tasks/shop-customer/Account/AddNewAddress.ts
```diff
@@ -0,0 +1,31 @@
...
+import type { Task } from '../../../types/Task';
+import type { FixtureTypes} from '../../../types/FixtureTypes';
+import { Address } from '../../../types/ShopwareTypes';
+
+export const AddNewAddress = base.extend<{ AddNewAddress: Task }, FixtureTypes>({
+    AddNewAddress: async ({ ShopCustomer, StorefrontAccountAddresses, StorefrontAccountAddressCreate }, use)=> {
+        const task = (address: Address) => {
+            return async function AddNewAddress() {
+
+                await ShopCustomer.goesTo(StorefrontAccountAddresses.url());
```
frobel: I would put that line (goesTo) into the test so that It is more transparent where you create the address.

frobel: and we save execution time because the fixture ShopCustomer is large and takes much more time to generate compared to a simple page object.

frobel (changes requested):

vanpham-sw:
The Address form changes a lot in 6.7 so it's actually not ready :(

src/services/TestDataService.ts
sguder: I think you have to move the customer_address in front of customer in this list, so it gets deleted first

vanpham-sw: turns out because my test need to clean up registered user separately because it needs a different registration flow, [see here](https://github.com/shopware/shopware/pull/6589/files#diff-cac930faa3e06ef7b7bdda15212073ce01218d16e6cd00961da01ac984ce66a7R46-R77)

src/types/ShopwareTypes.ts
```diff
@@ -111,8 +124,12 @@ export type Currency = components['schemas']['Currency'] & {
     id: string,
 }
 
-export type Country = components['schemas']['Country'] & {
+export type Country = Omit<components['schemas']['Country'], 'states'> & {
```
yusufttur, line 127: Why you're omitting states from the schema, and using your own custom states structure?

vanpham-sw: I don't need the entire structure of the states in the test, that's why I just include the needed fields. We can always add more later if we need other fields. For me, adding them now is overkill.

yusufttur (dismissed):

sguder (dismissed):

src/page-objects/storefront/AccountAddresssCreate.ts
```diff
@@ -12,17 +12,19 @@ export class AccountAddressCreate implements PageObject {
...
 
     constructor(public readonly page: Page) {
-        this.salutationDropdown = page.locator('#addresspersonalSalutation');
-        this.firstNameInput = page.locator('#addresspersonalFirstName');
-        this.lastNameInput = page.locator('#addresspersonalLastName');
-        this.companyInput = page.locator('#addresscompany');
-        this.departmentInput = page.locator('#addressdepartment');
-        this.streetInput = page.locator('#addressAddressStreet');
-        this.zipcodeInput = page.locator('#addressAddressZipcode');
+        this.salutationDropdown = page.getByRole('combobox', { name: 'Salutation' });
```
frobel, line 18: you checked if that change is also 6.7. and 6.6. compatible?

src/services/ShopwareDataHelpers.ts
```diff
@@ -286,4 +286,31 @@ export const getPromotionWithDiscount = async (promotionId: string, adminApiCont
 
     const { data: promotion } = (await resp.json()) as { data: Promotion[] };
     return promotion[0];
-};
\ No newline at end of file
+};
+
+export async function deleteRegisteredUser(adminApiContext: AdminApiContext, email: string): Promise<void> {
```
frobel: Just an idea to centralize the deletion process:

1. just search for a customer with an email (ex. TestDataService.getCustomer(email) )
2. use the returned customer.id and add it via TestDataService to the deletion queue (see step 3)
3. TestDataService.addCreatedRecord('customer', customerId);

src/page-objects/storefront/AccountAddresses.ts
```diff
@@ -1,19 +1,38 @@
...
     public readonly addNewAddressButton: Locator;
     public readonly editBillingAddressButton: Locator;
     public readonly editShippingAddressButton: Locator;
     public readonly useDefaultBillingAddressButton: Locator;
     public readonly useDefaultShippingAddressButton: Locator;
+    public readonly deliveryNotPossibleAlert: Locator | undefined;
+    public readonly availableAddresses: Locator | undefined;
+    public readonly addressDropdownButton: Locator | undefined
+    public readonly availableAddressesUseAsBillingAddress: Locator | undefined;
+    public readonly availableAddressesUseAsShippingAddress: Locator | undefined;
```
frobel, line 16: Is that working without error hints in IDE? I run also in same situation, that I have locators just available in 6.7 and not 6.6.

frobel (changes requested):

src/page-objects/storefront/AccountAddresses.ts
vanpham-sw: no errors
image

src/page-objects/storefront/AccountAddresssCreate.ts
vanpham-sw: Yes, I ran tests locally on 66 and 6.7

src/services/ShopwareDataHelpers.ts
vanpham-sw: Good suggestion! If it was there in the first place, I would save hours debugging the country data clean up

frobel (approved):

PR #301, opened 2025-02-05
feat: add page not found page

yusufttur:


src/page-objects/storefront/PageNotFound.ts
```diff
@@ -0,0 +1,20 @@
...
+
+    constructor(public readonly page: Page) {
+        this.pageNotFoundImage = page.getByAltText('Page not found');
+        this.headline = page.getByRole('heading', { name: 'Page not found' });
+        this.pageNotFoundMessage = page.getByText(`We are sorry, the page you're looking for could not be found. It may no longer exist or may have been moved.`);
+        this.backToShopButton = page.getByRole('link', { name: 'Back to shop' });
+    }
+
+    url() {
+        return 'non-existing-page';
```
frobel, line 18: Maybe you return an error message with:

`return new Error('Function not implemented, because it is a modal page object').message;`

yusufttur: Actually I need to visit this url to verify 404 page. see: https://github.com/shopware/shopware/pull/6593/files

frobel (dismissed):

frobel (dismissed):

frobel (approved):

Issue #305, opened 2025-02-05
Support to use the acceptance test suite on customized stores


DanieliMi:
This test suite is fantastic in that it already provides all the necessary fixtures, page objects, and other utilities needed to write tests for the store. However, it currently only works on clean installs with specific example (or default) data present. It would be incredibly useful if the suite allowed for more flexibility in configuring the default data being used. This would make it possible to run tests on customer projects that differ from a clean install.

### Example
For instance, we have developed a plugin for one of our customers and would like to test it in both testing and staging environments. Unfortunately, in these environments, we cannot use some of the admin page objects because the required shipping method "Standard" is missing. This shipping method is queried in the following part of the code: https://github.com/shopware/acceptance-test-suite/blob/19333ce285b77c045514614e368f60755cc42f6d/src/services/ShopwareDataHelpers.ts#L112-L127

To work around this, we currently have to completely overwrite some of the fixtures, which is not ideal. We would prefer to keep most of the suite logic intact and only override what is absolutely necessary.

### Solution
It would be great to introduce a mechanism that allows us to configure the default data used in the fixtures. Ideally, this could be done in one of the following ways:
* Let us somehow configure the default data to be used in the fixtures.
* A way to extend the fixtures or overwrite only certain properties instead of having to completely overwrite them.

I am aware of Playwright’s extension system but extending a fixture is not a viable solution in our example since the data is loaded when the fixture is created and it will be created if we inject the original fixture.

Would you be open to such a change?

pweyck:
First of all, thank you for the feedback 💙 

One of the original design goals for the package was that it "just works" with an `APP_URL` and auth credentials and that's only possible if we don't expect any particular state. So it's in full alignment with what you're asking for. We try to reach this goal by creating most data instead of relying on existing one. But obviously there's some global/sales channel level config, that might differ from use case to use case. 

We're still assuming some things about the shop:
- payment method is `InvoicePayment` 
- shipping method is `Standard` 
- `EUR` currency is available (system default is also added)
- `en-GB` language is available (system default is also added)
- `de` country is available
- you want to use the default `Storefront` theme

https://github.com/shopware/acceptance-test-suite/blob/ba108ea999e25e825c46ec308202b247d077c089/src/fixtures/DefaultSalesChannel.ts#L52-L59

If we define all of those as separate fixtures, you [could override](https://playwright.dev/docs/test-fixtures#overriding-fixtures) them individually. Would that solve your problem?

[...]

PR #308, opened 2025-02-06
feat: move sales channel default data to fixtures (#305)

DanieliMi:
Moving the logic from https://github.com/shopware/acceptance-test-suite/blob/ba108ea999e25e825c46ec308202b247d077c089/src/fixtures/DefaultSalesChannel.ts#L52-L59 to new fixtures within ShopwareDataFixtures for better overwriting of default data.

### DIFF
--- a/src/fixtures/ShopwareDataFixtures.ts
+++ b/src/fixtures/ShopwareDataFixtures.ts
@@ -1,0 +1,146 @@
+import { test as base } from '@playwright/test';
+import type { FixtureTypes } from '../types/FixtureTypes';
+import {
+    getCountryId,
+    getCurrency,
+    getLanguageData,
+    getPaymentMethodId,
+    getShippingMethodId,
+    getSnippetSetId,
+    getTaxId,
+    getThemeId,
+} from '../services/ShopwareDataHelpers';
+
+export interface Country {
+    id: string;
+}
+
+export interface Currency {
+    id: string;
+}
+
+export interface Language {
+    id: string;
+    translationCode: { id: string };
+}
+
+export interface PaymentMethod {
+    id: string;
+}
+
+export interface ShippingMethod {
+    id: string;
+}
+
+export interface SnippetSet {
+    id: string;
+}
+
+export interface Tax {
+    id: string;
+}
+
+export interface Theme {
+    id: string;
+}
+
+export interface ShopwareDataFixtureTypes {
+    Country: Country;
+    Currency: Currency;
+    Language: Language;
+    PaymentMethod: PaymentMethod;
+    ShippingMethod: ShippingMethod;
+    SnippetSet: SnippetSet;
+    Tax: Tax;
+    Theme: Theme;
+}
+
+export const test = base.extend<NonNullable<unknown>, FixtureTypes>({
+
+    Country: [
+        async ({ AdminApiContext }, use) => {
+            const countryId = await getCountryId('de', AdminApiContext);
+
+            await use({
+                id: countryId,
+            });
+        },
+        { scope: 'worker' },
+    ],
+
+    Currency: [
+        async ({ AdminApiContext }, use) => {
+            const currency = await getCurrency('EUR', AdminApiContext);
+
+            await use({
+                id: currency.id,
+            });
+        },
+        { scope: 'worker' },
+    ],
+
+    Language: [
+        async ({ AdminApiContext }, use) => {
+            const language = await getLanguageData('en-GB', AdminApiContext);
+
+            await use(language);
+        },
+        { scope: 'worker' },
+    ],
+
+    PaymentMethod: [
+        async ({ AdminApiContext }, use) => {
+            const paymentMethodId = await getPaymentMethodId(AdminApiContext);
+
+            await use({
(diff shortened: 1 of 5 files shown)


src/services/ShopwareDataHelpers.ts
```diff
@@ -105,27 +105,6 @@ export const getPaymentMethodId = async (adminApiContext: AdminApiContext, handl
     return result.data[0].id;
 };
 
-/**
- * Gives the default shipping method back called Standard
- * @param adminApiContext - An AdminApiContext entity
- */
-export const getDefaultShippingMethodId = async (adminApiContext: AdminApiContext): Promise<string> => {
```
pweyck, line 112: We should not removed it but just deprecate this method (otherwise it would be a break) and create the more general `getShippingMethodId`, that takes the name as the parameter.

src/fixtures/ShopwareDataFixtures.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export interface ShopwareDataFixtureTypes {
+    ShippingMethod: ShippingMethod;
+}
+
+export const test = base.extend<NonNullable<unknown>, FixtureTypes>({
+
+    ShippingMethod: [
+        async ({ AdminApiContext }, use) => {
+            const resp = await AdminApiContext.post('search/shipping-method', {
```
pweyck: Call the new method with `'Standard'` here

src/fixtures/ShopwareDataFixtures.ts
```diff
@@ -0,0 +1,36 @@
...
+    id: string;
+}
+
+export interface ShopwareDataFixtureTypes {
+    ShippingMethod: ShippingMethod;
+}
+
+export const test = base.extend<NonNullable<unknown>, FixtureTypes>({
+
+    ShippingMethod: [
```
pweyck, line 102: For `PaymentMethod` we could do the same

pweyck (changes requested):
In general we should keep the helper functions, because they are useful in other contexts, but discourage (and eventually remove) the use of the ones which assume some data exists like `getDefaultShippingMethodId`

DanieliMi:
Thank you for having a look and the feedback!

> In general we should keep the helper functions, because they are useful in other contexts, but discourage (and eventually remove) the use of the ones which assume some data exists like `getDefaultShippingMethodId`

I just had a deeper look in how the helpers are used otherwise (I just quickly scanned it before when I made the suggestion) and I do agree. Most of them are used during data creation and the methods have parameters with which you  can easily provide the data you want. I'll then only change the ones that provide fixed default data in a fixture.

src/services/ShopwareDataHelpers.ts
DanieliMi: Good point! I will change that and even use it in the fixture. Since it is a single fixture which we can override independently I see no problem in using the helper.

src/fixtures/ShopwareDataFixtures.ts
DanieliMi: Yes, I intend to change all of https://github.com/shopware/acceptance-test-suite/blob/ba108ea999e25e825c46ec308202b247d077c089/src/fixtures/DefaultSalesChannel.ts#L52-L59
to an fixture. Then you can overwrite them as needed.

DanieliMi:
Hi @pweyck, I've done the requested changes and also moved the remaining default data logic. Can you have a look again?

frobel (approved):

sguder (approved):

DanieliMi:
Do you have any idea why the two tests fail?

frobel:
> Do you have any idea why the two tests fail?

1. Your second commit message is not using our conventional commit policy. 
2. I will rerun the pipeline. Maybe there was a temporarily issue in the API.

pweyck (approved):

DanieliMi:
> > Do you have any idea why the two tests fail?
> 
> 1. Your second commit message is not using our conventional commit policy.
> 2. I will rerun the pipeline. Maybe there was a temporarily issue in the API.

@frobel I presume you mean b4415946986efd11c58d05e2bff30876d47d63e4. Can you explain to me what does match the conventional commit policy so I can fix it? I have read it and thought it does comply. The other commit is just a merge commit from updating the branch via GitHub directly.

pweyck:
> > > Do you have any idea why the two tests fail?
> > 
> > 
> > 
> > 1. Your second commit message is not using our conventional commit policy.
> > 2. I will rerun the pipeline. Maybe there was a temporarily issue in the API.
> 
> @frobel I presume you mean [b441594](https://github.com/shopware/acceptance-test-suite/commit/b4415946986efd11c58d05e2bff30876d47d63e4). Can you explain to me what does match the conventional commit policy so I can fix it? I have read it and thought it does comply. The other commit is just a merge commit from updating the branch via GitHub directly.

I've used squash merge, in that case it does not matter, because only the title is used for the commit message

frobel:
@DanieliMi we released your change in the ATS version 11.9.0

DanieliMi:
Ah alright :)  
Thank you very much!

PR #309, opened 2025-02-06
fix: make contact form locators compatible with 6.7. and 6.6.

frobel:


src/page-objects/storefront/ContactForm.ts
```diff
@@ -27,28 +38,53 @@ export class ContactForm extends Home implements PageObject {
...
+            this.subjectInput = this.contactCmsPage.getByLabel('Subject');
+            this.commentInput = this.contactCmsPage.getByLabel('Comment');
+            this.privacyPolicyCheckbox = this.contactCmsPage.getByRole('checkbox', { name: 'By selecting continue you confirm that you have read and agree to our' });
+            this.submitButton = this.contactCmsPage.getByRole('button', { name: 'Submit' });
+            this.cardTitle = this.contactCmsPage.locator('.card-title');
+            this.basicCaptcha = this.contactCmsPage.locator('.basic-captcha');
+            this.greCaptchaV2Container = this.contactCmsPage.locator('.grecaptcha-v2-container');
+            this.greCaptchaV2Input = this.contactCmsPage.locator('.grecaptcha-v2-input');
+            this.greCaptchaProtectionInformation = this.contactCmsPage.locator('.grecaptcha-protection-information');
+        }
```
vanpham-sw: Just thinking without testing, can we shorten this by doing like this:

```
if (satisfies(instanceMeta.version, '<6.7')) {
    this.contactWrapper = this.page.getByRole('dialog').filter({ has: this.page.getByText('Contact') });
} else {
    this.contactWrapper = this.page.locator('.card').filter({ has: this.page.getByText('Contact') });
}

this.basicCaptcha = this.contactWrapper.locator('.basic-captcha');
......other common locators

```

frobel: yes should work. thx for the hint. Would you then deprecate the contactModal property or would you go with two properties with the same behaviour? So: contactWrapper and contactModal?

vanpham-sw: I would deprecate the contactModal property, i guess this is possible?

```
public readonly contactWrapper: Locator;

/**
 * @deprecated Compatible until 6.6.x, will be removed in 6.8.x, use 'contactWrapper' instead
 */
public get contactModal(): Locator {
    console.warn("contactModal is deprecated and will be removed in version 6.8.x. Please use contactWrapper instead.");
    return this. contactWrapper;
}
```

frobel: adjusted

vanpham-sw (approved):

PR #315, opened 2025-02-12
fix: change name to getCountry because method return country object

vanpham-sw:


src/services/TestDataService.ts
```diff
@@ -1775,11 +1775,11 @@ export class TestDataService {
     };
 
     /**
-     * Retrieves a country Id based on its iso2 code.
+     * Retrieves a country based on its iso2 code.
      *
      * @param iso2 - The iso2 code of the country, for example "DE".
      */
-    async getCountryId(iso2: string): Promise<Country> {
+    async getCountry(iso2: string): Promise<Country> {
```
frobel, line 1792: That would be a breaking change. I suggest deprecating the method and creating a new one with the name you suggested but with the same behaviour.

After collecting some deprecations, we can release a new major and remove all collected deprecations.

vanpham-sw: Thx for the suggestion, updated to remove breaking change

frobel (approved):

PR #322, opened 2025-02-18
Plugin config helper

mdanierov:


frobel:
Hi @mdanierov, thanks for your contribution. Could you please shortly describe what use-case you want to cover? I just thinking about our TestDataService which also contains the possibility to change the SystemConfig values via [TestDataService.setSystemConfig()](https://github.com/shopware/acceptance-test-suite/blob/49d8aecb434252e76670c35a183d33cc20c650e9/src/services/TestDataService.ts#L817) for a specific SalesChannel. So it is kind of "worker" related because the SalesChannel, which is set by default, is worker-related.

frobel:
Hi @mdanierov we will close your request due to a lack of response. We wanna keep our backlog as clean and up-to-date as possible. Feel free to create a new Issue with your suggestion again.

PR #334, opened 2025-02-25
fix: adjust OrderCreation.spec.js in order not to update write-protected line-item properties, increase minimum version for feature toggle route and adjust integration test routine

### LINKED ISSUE #313: Update build process for E2E test images
- Implement regular build process, the images are stale at the moment
    - also, move this workflow into the ATS repo; currently this is located in shopware/shopware, but never used there (**TBD**)
- Use current `6.5.x` image build as a baseline, making use of the deployment helper everywhere

philipreinken:
chore(ci): update compose.yaml
chore(ci): add shopware.Dockerfile for test image builds
chore(ci): adjust test workflow

Refs: https://github.com/shopware/acceptance-test-suite/issues/313


compose.yaml
```diff
@@ -45,50 +38,78 @@ x-shopware: &shopware-65
...
+  init:
+    <<: [ *service, *shopware-volumes ]
+    image: ghcr.io/shopware/acceptance-test-suite/test-image:${SHOPWARE_VERSION:-trunk}
+    build:
+      context: .
+      dockerfile: test-image.Dockerfile
+      args:
+        - PHP_VERSION=${PHP_VERSION:-8.4}
+        - SHOPWARE_VERSION=${SHOPWARE_VERSION:-trunk}
+        - SHOPWARE_BUILD_SOURCE=${SHOPWARE_BUILD_SOURCE:-branch}
```
pweyck, line 90:
```suggestion
```
The images are build in the test jobs right. We should only build them in the nightly

test-image.Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+
+ARG SHOPWARE_VERSION
+
+RUN shopware-cli project create /src ${SHOPWARE_VERSION#v} --verbose
+
+FROM shopware-cli AS checkout-branch
+
+ARG SHOPWARE_VERSION
+
+ENV COMPOSER_ROOT_VERSION="6.7.9999999-dev"
```
pweyck: this should not be necessary, except for trunk. For other versions we should use the tags

.github/workflows/build-shopware-image.yml
```diff
@@ -0,0 +1,49 @@
+name: Build Shopware Image
+
+on:
+  schedule:
+    - cron: '0 1 * * *'
+  workflow_dispatch:
+
+jobs:
+  build-test-image:
```
pweyck, line 22: we should build the image for different released versions addition to 6.6.x and trunk branch. You can use the action [shopware/github-actions/versions@main](https://github.com/shopware/shopware/blob/e3c4e85f881e18141bfb87a9ff261afb33e6e5a0/.github/workflows/integration.yml#L330-L352)

The ats should only test against released versions not trunk (because it should continue to work with older versions, otherwise it's pretty useless for externals)

compose.yaml
philipreinken: This is useful for regular usage of the `compose.yaml`, so I'd like to keep it; once the images have been registered/pushed by the nightly once, they will not be built anymore, but only pulled by the test job.

test-image.Dockerfile
philipreinken: This only comes into play when checking out a development branch (for example trunk). [Tags are set up](https://github.com/shopware/acceptance-test-suite/pull/334/files/4a1a21e26f482dcab172a46786c00a8017cd2def#diff-01f0199904147627781673f29653c3c3430acb09b5cbbe5d2f5d454e13c6e6eeR11) using `shopware-cli project create [version]`.

But I just realised that `6.6.x` will also use this value in the current state, so I could make this configurable as well. Sounds alright?

.github/workflows/build-shopware-image.yml
philipreinken: Makes sense; I'll add those test runs 👍

src/fixtures/HelperFixtures.ts
```diff
@@ -68,7 +68,7 @@ export const test = base.extend<NonNullable<unknown>, FixtureTypes>({
             const config = (await response.json()) as { version: string };
 
             const features: FeaturesType = {};
-            if (satisfies(config.version, '>=6.6.0.0')) {
+            if (satisfies(config.version, '>=6.6.1.0')) {
```
philipreinken, line 71: @pweyck it looks like the `/_action/feature-flag` route was introduced with `v6.6.1.0` and is not present in `v6.6.0.0`.

philipreinken: Not sure if I break something by increasing this, but it will fail otherwise when executed with `v6.6.0.0`.

philipreinken:
Adjustments to the required checks: https://github.com/shopware/organization/pull/295

pweyck (approved):

compose.yaml
```diff
@@ -1,42 +1,35 @@
 x-env: &env
```
pweyck, line 1: I think we should also update the documentation on how to run it locally

philipreinken: I've adjusted the docs 😊

pweyck (approved):

.github/workflows/build-shopware-image.yml
```diff
@@ -0,0 +1,74 @@
...
+
+jobs:
+  versions:
+    name: Determine Shopware versions to build
+    runs-on: ubuntu-24.04
+    outputs:
+      first-version: ${{ steps.versions.outputs.first-version }}
+      latest-version: ${{ steps.versions.outputs.latest-version }}
+      lts-first-version: ${{ steps.versions.outputs.lts-first-version }}
+      lts-latest-version: ${{ steps.versions.outputs.lts-latest-version }}
```
yusufttur: Can you explain a bit what these versions mean?

philipreinken: Ah yes, I've just copied those over, I should have added some comments:
- `first`: First version of the current major, e.g. `v6.6.0.0`
- `latest`: Latest released version of the current major, e.g. `v6.6.10.1`
- `lts-first`: First version of the previous major, e.g. `v6.5.0.0`
- `lts-latest`: Latest version of the previous major, e.g. `v6.5.8.16`

yusufttur (approved):

PR #337, opened 2025-02-28
fix: always check storefront availability after theme assign

pweyck:
Due to new async caching/invalidation behavior it's possible that the storefront is not yet fully ready after the theme assignment, even if it's synchronous. So we should always check if the theme is loaded before going forward.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -84,7 +84,7 @@ jobs:
     - name: Run Playwright tests against Shopware ${{ matrix.SHOPWARE_VERSION }}
       env:
         APP_URL: http://localhost:${{ matrix.PORT }}
-      run: npx playwright test --repeat-each 1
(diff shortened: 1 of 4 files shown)


yusufttur (approved):

philipreinken (approved):

.github/workflows/test.yml
```diff
@@ -84,7 +84,7 @@ jobs:
     - name: Run Playwright tests against Shopware ${{ matrix.SHOPWARE_VERSION }}
       env:
         APP_URL: http://localhost:${{ matrix.PORT }}
-      run: npx playwright test --repeat-each 4
+      run: npx playwright test --repeat-each 1
```
pweyck:
```suggestion
      run: npx playwright test --repeat-each 4
```

frobel (approved):

PR #345, opened 2025-03-03
fix: use delayed clear

pweyck:


src/services/TestDataService.ts
```diff
@@ -2521,7 +2522,7 @@ export class TestDataService {
     }
 
     async clearCaches() {
-        await this.AdminApiClient.delete('_action/cache');
+        await clearDelayedCache(this.AdminApiClient);
```
pweyck, line 2525: this might only work in 6.7

PR #364, opened 2025-03-24
feat: bump api-client version to 1.2.1

frobel:


tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "target": "ES6",
     "lib": ["ESNext"],
     "module": "ESNext",
-    "moduleResolution": "node",
+    "moduleResolution": "Bundler",
```
frobel, line 6: @pweyck is that change of the moduleResolver a breaking change? And would it somehow affect the compatibility with our platform commercial, etc, repos?

frobel: @pweyck do you have any objections/opinion here?

pweyck: We're already using a bundler, so I guess it's ok. If we can update this without changing any config in the projects that use it, I'd say this not breaking

pweyck (approved):

pweyck (approved):

PR #374, opened 2025-04-24
feat: FlowValidation test requirements

### LINKED ISSUE #8352: ATS: Flow Builder Validation
As an admin user, I want that certain actions get executed based on the flow, so that I can automate the processes.

{noformat}GIVEN there is a flow
AND this flow starts with a condition
AND there are actions for the True (action1) and False (action2) branch of the condition
AND there is an additional action3, that is not bound to a condition
WHEN the trigger of the flow gets triggered
THEN based on the <outcome> of the condition the <condition_based_action> will be executed
AND action3 will be executed
AND it can be validated, that the correct actions have been executed{noformat}

||*outcome*||*conditon_based_action*||
|true|action1|
|false|action2|

GitEvil:
https://github.com/shopware/shopware/pull/8784


GitEvil:
Issue: [#8352](https://github.com/shopware/shopware/issues/8352)

yusufttur (approved):

PR #375, opened 2025-04-25
feat: add header and footer pages

sguder:


package-lock.json
yusufttur, line 1: Do you need to update all the dependencies here? Otherwise, it shouldn't be part of this PR, am I right? 🤔

yusufttur (changes requested):

package-lock.json
sguder: Honestly don't know when to do it and when not. But I wasn't able to compile without doing it..

yusufttur (approved):

PR #384, opened 2025-05-12
fix: increase SaaS timeout to 90sec

frobel:


src/fixtures/PageContexts.ts
```diff
@@ -101,7 +101,7 @@ export const test = base.extend<FixtureTypes>({
             return url.hash !== '#login';
         });
 
-        await expect(page.getByText('Administration').first()).toBeVisible();
+        await expect(page.getByText('Administration').first()).toBeVisible({ timeout: 120000 });
```
vanpham-sw: It looks to me that the new 120 timeout will affect not only sass test, cmiiw

src/fixtures/PageContexts.ts
```diff
@@ -58,7 +58,7 @@ export const test = base.extend<FixtureTypes>({
 
         await expect(page.url()).toContain('login');
 
-        await page.getByLabel(/Username|Email address/).fill(adminUser.username);
+        await page.getByLabel(/Username|Email address/).fill(adminUser.username, { timeout: 120000});
```
vanpham-sw: It looks to me that the new 120 timeout will affect not only sass test, cmiiw

src/fixtures/PageContexts.ts
frobel: I'm still checking what the root problem is. I guess it is the initial login page loading, which takes very long the first time.

src/fixtures/PageContexts.ts
frobel: I'm still checking what the root problem is. I guess it is the initial login page loading, which takes very long the first time.

src/fixtures/PageContexts.ts
```diff
@@ -57,6 +57,7 @@ export const test = base.extend<FixtureTypes>({
         });
 
         await expect(page.url()).toContain('login');
+        await expect(page.getByLabel(/Username|Email address/)).toBeVisible({ timeout: 120000 });
```
frobel: The timeout is necessary here because the Login Page takes some time to load all resources to the cache.

src/fixtures/PageContexts.ts
```diff
@@ -100,8 +101,7 @@ export const test = base.extend<FixtureTypes>({
         await page.waitForURL((url) => {
             return url.hash !== '#login';
         });
-
-        await expect(page.getByText('Administration').first()).toBeVisible();
+        await expect(page.getByText('Administrator').first()).toBeVisible();
```
frobel: I changed it to "Administrator" because in Cloud we don't have a hint in the top left corner which called "Administration". Therefore, I chose the term "Administrator" which is shown in the left bottom corner of the user profil.

kstala (approved):

Issue #392, opened 2025-05-18
dotenv app_url wrong: "https://127.0.0.1api/"


ic-msc:
Shopware 6 /.env:
APP_URL=https://127.0.0.1

shopware/ATS is using: 
```javascript
  static async createApiRequestContext(options) {
    const extraHTTPHeaders = {
      "Accept": "application/json",
      "Content-Type": "application/json"
    };
    if (options.access_token && options.access_token.length) {
      extraHTTPHeaders["Authorization"] = "Bearer " + options.access_token;
    }
    return await request.newContext({
      baseURL: `${options.app_url}api/`,
      ignoreHTTPSErrors: options.ignoreHTTPSErrors,
      extraHTTPHeaders
    });
  }
``` 

So you're creating the baseURL as "https://127.0.0.1api/" – The default is to NOT add a trailing slash in the shopware 6 .env file.

Also you're ignoring the Playwright BaseURL.

Something's very strange here.

pweyck:
Playwright needs the slash at the end, so that relative urls work. We handle that in the playwright config of our projects. Example https://github.com/shopware/shopware/blob/ce8551ba246a841fb9e6f5480aaa46e72ff0e6b3/tests/acceptance/playwright.config.ts#L35-L40

I'm not sure, if we can fix this in the library, but we'll try to find a solution

jc-sw-git:
Decided to move it to the Prod Operations domain, after a brief chat with @pweyck who has already worked on it

PR #394, opened 2025-05-21
feat!: refactor configuration (fixes: #392)

pweyck:
- Introduces new `ConfigOptions.ts` file for centralized Shopware configuration
- Removes strict requirement for URLs to end with a slash
- Adds type safety to configuration system

TODO:
- update docs

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -1,31 +1,11 @@
 import { defineConfig, devices } from '@playwright/test';
+import { ConfigOptions } from './src/ConfigOptions';
 
-process.env['SHOPWARE_ADMIN_USERNAME'] = process.env['SHOPWARE_ADMIN_USERNAME'] || 'admin';
-process.env['SHOPWARE_ADMIN_PASSWORD'] = process.env['SHOPWARE_ADMIN_PASSWORD'] || 'shopware';
-process.env['MAILPIT_BASE_URL'] = process.env['MAILPIT_BASE_URL'] || 'http://localhost:8013';
+const defaultAppURL = 'http://localhost:8011';
+const appURL = process.env['APP_URL'] ?? defaultAppURL;
+const defaultCommand = appURL === defaultAppURL ? 'docker compose up --pull=always --quiet-pull' : 'sleep 999d';
 
-const defaultAppUrl = 'http://localhost:8011/';
-process.env['APP_URL'] = process.env['APP_URL'] ?? defaultAppUrl;
-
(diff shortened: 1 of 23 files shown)


pweyck:
I'm considering refactoring it to https://playwright.dev/docs/test-parameterize#parameterized-projects, this would get rid of all custom env vars (we'll still use them as the default) and we get auto completion in the config.

sguder (approved):

playwright.config.ts
```diff
@@ -1,31 +1,11 @@
 import { defineConfig, devices } from '@playwright/test';
+import { ConfigOptions } from './src/ConfigOptions';
```
frobel, line 2: We have to adjust or at least inform people who wrote the [following documentation](https://hub.shopware.com/learn/unit/playwright-end-to-end-tests) in hub.shopware.com

pweyck: All old playwright.config.ts should still work after this change (and benefit from the normalization/slash handling). We could recommend the new typed way though

src/services/AdminApiContext.ts
```diff
@@ -36,9 +31,8 @@ export class AdminApiContext {
         this.options = options;
     }
 
-    public static async create(options?: AdminApiContextOptions) {
+    public static async create(options: AdminApiContextOptions) {
```
pweyck, line 34: strictly speaking, this is a break

tests/PageObjects/BaseConfig.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+    expect(BaseConfig.shopware.adminAPIURL).toBe(shopware.adminAPIURL);
+
+    // check that all URLs work
+    const pageContext = await browser.newContext({
+        baseURL: BaseConfig.shopware.appURL,
+    });
+    const page = await pageContext.newPage();
+    await page.goto(BaseConfig.shopware.appURL);
+    await page.goto(BaseConfig.shopware.adminURL);
+    await page.goto(BaseConfig.shopware.adminAPIURL);
```
frobel, line 23: Actually that not look like a test:

suggestion:

```
const urls = [
    BaseConfig.shopware.appURL,
    BaseConfig.shopware.adminURL,
    BaseConfig.shopware.adminAPIURL
  ];

  for (const url of urls) {
    const response = await page.goto(url);
    expect(response?.ok(), `Failed to load: ${url}`).toBeTruthy();
  }
```

PR #401, opened 2025-06-04
chore: add possibility to completely disable cleanup functionality

philipreinken:
Refs: https://github.com/shopware/shopware/issues/9907#issuecomment-2922429009

### DIFF
--- a/src/fixtures/TestData.ts
+++ b/src/fixtures/TestData.ts
@@ -2,6 +2,9 @@ import { test as base } from '@playwright/test';
 import { TestDataService } from '../services/TestDataService';
 import type { FixtureTypes } from '../types/FixtureTypes';
 
+const ATS_SKIP_CLEANUP: boolean = ['1', 'true'].includes(process.env['ATS_SKIP_CLEANUP'] || '');
+const ATS_EXEC_CLEANUP = !ATS_SKIP_CLEANUP;
+
 export interface TestDataFixtureTypes {
     TestDataService: TestDataService;
 }
@@ -21,6 +24,8 @@ export const test = base.extend<FixtureTypes>({
 
         await use(DataService);
 
-        await DataService.cleanUp();
+        if (ATS_EXEC_CLEANUP) {
+            await DataService.cleanUp();
+        }
     },
 });


frobel:
Hi @philipreinken I just want to mention, that we have also a test which is using the CleanUp and playing around with it:

see [here](https://github.com/shopware/acceptance-test-suite/blob/9fc405ce42c96dbcbe1f80ef2a8c4977f2244518/tests/TestDataService/TestDataService.spec.ts#L108)

or [here](https://github.com/shopware/acceptance-test-suite/blob/9fc405ce42c96dbcbe1f80ef2a8c4977f2244518/tests/TestDataService/SystemConfig.spec.ts#L37)

or [here](https://github.com/shopware/acceptance-test-suite/blob/9fc405ce42c96dbcbe1f80ef2a8c4977f2244518/tests/TestDataService/OrderCreation.spec.ts#L76)

src/services/TestDataService.ts
```diff
@@ -1692,6 +1692,10 @@ export class TestDataService {
      * @param shouldCleanUp - The config setting for the automated data clean up. Default is "true".
      */
     setCleanUp(shouldCleanUp = true) {
+        if (process.env['ATS_SKIP_CLEANUP'] && shouldCleanUp) {
```
pweyck: I'd put this in the fixture instead

philipreinken:
> Hi @philipreinken I just want to mention, that we have also a test which is using the CleanUp and playing around with it:
> 
> see [here](https://github.com/shopware/acceptance-test-suite/blob/9fc405ce42c96dbcbe1f80ef2a8c4977f2244518/tests/TestDataService/TestDataService.spec.ts#L108)
> 
> or [here](https://github.com/shopware/acceptance-test-suite/blob/9fc405ce42c96dbcbe1f80ef2a8c4977f2244518/tests/TestDataService/SystemConfig.spec.ts#L37)
> 
> or [here](https://github.com/shopware/acceptance-test-suite/blob/9fc405ce42c96dbcbe1f80ef2a8c4977f2244518/tests/TestDataService/OrderCreation.spec.ts#L76)

Thanks @frobel ! I've seen those instances, but I'd like to be able to override this behaviour globally. When just calling `setCleanup(false)` any other test could also call `setCleanup(true)` and everything's back to normal again, if I'm not mistaken.

src/services/TestDataService.ts
philipreinken: Should be resolved in the latest revision 😊

frobel (approved):

pweyck (approved):

PR #402, opened 2025-06-04
chore: execute changed test files repeatedly

### LINKED ISSUE #398: Add repeated execution for added/changed tests
Might be the case already, we could think about increasing the number of repetitions / only repeat added & changed tests instead of _all_ tests.

philipreinken:
Resolves: https://github.com/shopware/acceptance-test-suite/issues/398

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -66,6 +66,12 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
+    - id: changed-test-files
+      uses: tj-actions/changed-files@ed68ef82c095e0d48ec87eccea555d944a631a4c # v46
+      with:
+        files: |
+          tests/**/*.spec.ts
+
     - name: Login into Github Container Registry
       run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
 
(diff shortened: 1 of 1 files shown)


frobel:
Hi @philipreinken maybe there is a misunderstanding. We want to have that repeat function in the commercial and platform repository so that we are sure, changes in our test files are not flaky afterwards. 

But of course we can also implement it in that repository, too. :)

frobel (approved):

philipreinken:
> Hi @philipreinken maybe there is a misunderstanding. We want to have that repeat function in the commercial and platform repository so that we are sure, changes in our test files are not flaky afterwards.
> 
> But of course we can also implement it in that repository, too. :)

Got it :+1: I've implemented a similar change in platform already, for commercial there's a PR open as well: https://github.com/shopware/shopware/issues/9907 (👀 sub-issues)

I just noticed something I didn't address in this PR though - currently all tests are repeated 4 times by default. Additionally, with this PR, all changed files will be repeated 10 times.
@frobel Do you think it makes sense to remove the 4x repetition by default and only keep the 10x for changed files? (as I don't want to bloat execution time too much)

pweyck:
> I just noticed something I didn't address in this PR though - currently all tests are repeated 4 times by default. Additionally, with this PR, all changed files will be repeated 10 times.
> @frobel Do you think it makes sense to remove the 4x repetition by default and only keep the 10x for changed files?

I'm not @frobel but I've added the repeated execution. I think we should keep it because every change in this library can have a big impact. I think its still fast enough, but if it gets to slow we can also just add shards instead of reducing `--repeat-each`

pweyck (approved):

frobel:
Yes, if the execution time is not overwhelming us, then I'm also fine to keep the 4x times execution.

PR #410, opened 2025-07-01
feat: a11y poc

### LINKED ISSUE #8737: Update RegisteredUserBuysProduct.spec.ts to use Keyboard only
### Technical TODO

Adapt the RegisteredUserBuysProduct.spec.ts to only navigate and interact with the storefront using a keyboard. You do not need to include the storefront skip links and may begin with the email input - as it is currently written.

### LINKED ISSUE #10890: Keyboard with Tabbing
### Technical TODO

Adapt the RegisteredUserBuysProduct.spec.ts to only navigate and interact with the storefront using a keyboard and include tabbing between elements

### LINKED ISSUE #10891: Custom Assertion: toHaveVisibleFocus
### Technical TODO

Create a custom assertion to check for visible focus indicators in the ATS

### LINKED ISSUE #10892: Keyboard with Bundled Assertions
### Technical TODO

Adapt the RegisteredUserBuysProduct.spec.ts to only navigate and interact with the storefront using a keyboard, including the [custom assertion: toHaveVisibleFocus](https://github.com/shopware/shopware/issues/10891), and bundled with a collection of assertions to cover missing [actionability checks](https://playwright.dev/docs/actionability#introduction)

ravenokko:
This is a proof of concept to extend the ATS to include additional support for accessibility. With an initial focus on the storefront, we can update tests and tasks to:

- only navigate and interact with shopware using a keyboard
- check for visible focus indicators as playwright focuses each interactive element

The following tests have been conducted to assert best practices and ensure no flakiness will be introduced:

- [x] Adapt tasks for RegisteredUserBuysProduct.spec.ts to only navigate and interact with the storefront using a keyboard (does not need to include the storefront skip links)
- [x] Adapt tasks for RegisteredUserBuysProduct.spec.ts to only navigate and interact with the storefront using a keyboard and include tabbing between elements (concluded to be impractical)
- [x] Create a custom assertion to check for visible focus indicators in the ATS: `toHaveVisibleFocus()`
- [x] Adapt tasks for RegisteredUserBuysProduct.spec.ts to only navigate and interact with the storefront using a keyboard, including the Custom Assertion: `toHaveVisibleFocus()`
- [x] Benchmark failure rate, runtime, and stability for all storefront-facing tests and tasks onPrem and in Cloud, with and without additional [actionability checks](https://playwright.dev/docs/actionability)


src/fixtures/a11y_poc/a11y_assertions/fooBar.ts
```diff
@@ -0,0 +1,36 @@
...
+    const expected = 'rgb(0, 128, 0)';
+    const pass = received === expected;
+    return {
+      message: () =>
+        pass
+          ? `raven expected ${received} not to be green`
+          : `raven expected ${received} to be green`,
+      pass,
+    };
+  }
```
ravenokko:
```suggestion
  },
```

src/fixtures/CustomMatchers.ts
```diff
@@ -0,0 +1,52 @@
...
+      if (!boxShadowStyle && !borderStyle && !outlineStyle) {
+        return {
+          message: () => `expected ${locator} to have visible focus, but no outline, border or box-shadow was detected`,
+          pass: false,
+        };
+      }
+
+      return {
+        message: () => `expected ${locator} to have visible focus`,
+        pass: true,
```
frobel, line 41: Why is it `true` here?

src/fixtures/CustomMatchers.ts
```diff
@@ -0,0 +1,52 @@
+import { expect as baseExpect, Locator } from '@playwright/test';
```
frobel: Could you import Locator as type?

```suggestion
import type { Locator } from '@playwright/test';
```

src/fixtures/CustomMatchers.ts
ravenokko: I think because it is distinguishing between 2 fail cases:

1. Failed because of no visible focus indicator
2. Failed for some other reason (404, server error, etc.)

@kstala can you maybe give any more details?

src/fixtures/CustomMatchers.ts
ravenokko: Done

src/fixtures/CustomMatchers.ts
```diff
@@ -0,0 +1,52 @@
+import { expect as baseExpect } from '@playwright/test';
+import type { Locator } from '@playwright/test';
```
ravenokko:
```suggestion
import type { Locator } from 'playwright-core';
```

src/services/Actor.ts
```diff
@@ -1,5 +1,5 @@
 import { test, expect } from '@playwright/test';
-import type { Page } from 'playwright-core';
+import type { Page, Locator } from '@playwright/test';
```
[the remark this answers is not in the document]
ravenokko:
```suggestion
import type { Page, Locator } from 'playwright-core';
```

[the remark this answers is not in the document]
frobel: I guess we changed it because of the strict mode we should follow. The pipeline was complaning about it 2 month before :see_no_evil:

src/fixtures/CustomMatchers.ts
frobel: I guess we changed it because of the strict mode we should follow. The pipeline was complaning about it 2 month before 🙈

yusufttur (approved):

src/fixtures/CustomMatchers.ts
ravenokko: It is how it is written in the Playwright documentation: https://playwright.dev/docs/test-assertions#add-custom-matchers-using-expectextend

vanpham-sw (approved):
Do we need a documentation PR somewhere?

src/index.ts
```diff
@@ -13,6 +13,8 @@ import { test as ShopCustomerTasks } from './tasks/shop-customer-tasks';
 import { test as FeatureService } from './fixtures/Feature';
 import { test as ShopwareDataFixture } from './fixtures/ShopwareDataFixtures';
 
+import './fixtures/CustomMatchers';
```
frobel, line 16: TShould we not import it as `{ Custom Matcher }` and export is also to make it a11ble for other developers? :thinking:

ravenokko: The problem is the default code for `mergeExpects` - as per the [official documentation to combine multiple custom matchers](https://playwright.dev/docs/test-assertions#combine-custom-matchers-from-multiple-modules) - doesn't work out of the box with our ATS. If you try, it throws a lot of errors and seems to want us to declare our own explicit type (which we shouldn't have to do).

My best guess is the fact we already override `expect` with `expects` via our Actor pattern is potentially part of the source of Playwright's confusion, but I unfortunately wasn't able to find a solution beyond this one.

But even though it looks a bit different, it is possible in tests to use `toHaveVisibleFocus()` as its own Locator assertion with or without the Actor pattern.

`await ShopCustomer.expects(StorefrontCheckoutConfirm.paymentMethod).toHaveVisibleFocus();`
`await expect(StorefrontCheckoutConfirm.paymentMethod).toHaveVisibleFocus();`

ravenokko:
> Do we need a documentation PR somewhere?

I have a separate issue for handling that: https://github.com/shopware/shopware/issues/8767

PR #413, opened 2025-07-02
feat: add HideElementsForScreenshot helper

ocavli:
This PR introduces a small refactoring to improve the readability and maintainability of our visual tests.

Extracted the inline addStyleTag call used for hiding dynamic or irrelevant UI elements into a reusable helper method: hideElementsForScreenshot.

The helper accepts a page instance and an array of selectors to hide using display: none !important.

This makes it easier for others—especially those less familiar with CSS—to understand what the code does without having to parse raw CSS.

Also improves consistency across test files by avoiding repetition.

Thanks @vanpham-sw  for the suggestion — wrapping this into a method definitely improves clarity 🙌

SEE COMMENT: https://github.com/shopware/shopware/pull/9771/files#r2168684991

### DIFF
--- a/src/fixtures/HelperFixtures.ts
+++ b/src/fixtures/HelperFixtures.ts
@@ -5,6 +5,7 @@ import type { FixtureTypes } from '../types/FixtureTypes';
 import { getCurrency, getLanguageData } from '../services/ShopwareDataHelpers';
 import { AdminApiContext } from '../services/AdminApiContext';
 import { satisfies } from 'compare-versions';
+import type { Page } from '@playwright/test';
 
 type FeaturesType = Record<string, boolean>;
 
@@ -16,6 +17,7 @@ export interface HelperFixtureTypes {
         isSaaS: boolean,
         features: FeaturesType,
     },
+    HideElementsForScreenshot: (page: Page, selectors: string[]) => Promise<void>
 }
 
 export const test = base.extend<NonNullable<unknown>, FixtureTypes>({
@@ -85,4 +87,21 @@ export const test = base.extend<NonNullable<unknown>, FixtureTypes>({
         },
         { scope: 'worker' },
     ],
+
+    HideElementsForScreenshot: [
+        async ({ }, use) => {
+        const fn = async (page: Page, selectors: string[]) => {
+            if (!selectors.length) return;
+
+            const css = selectors
+            .map(selector => `${selector} { display: none !important; }`)
+            .join('\n');
+
+            await page.addStyleTag({ content: css });
+        };
+
+        await use(fn);
+        },
+        { scope: 'worker' },
+  ],
 });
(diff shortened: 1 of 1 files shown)


vanpham-sw (approved):

src/fixtures/HelperFixtures.ts
```diff
@@ -85,4 +87,21 @@ export const test = base.extend<NonNullable<unknown>, FixtureTypes>({
         },
         { scope: 'worker' },
     ],
-});
+
+    HideElementsForScreenshot: [
+        async ({ }, use) => {
```
frobel, line 92: We have here the possibility to inject fixtures, so we could think about create a method with injected Page object on an AdminPage and another with StorefrontPage. You already thought about it?  Of cause that method makes it more flexible :)

ocavli: If I understood you correctly, you mean adding this method to every page object we use. 
Wouldn’t that cause a lot of code duplication, because we are using it for most of the admin pages currently? With the current approach, we can use it directly as a general helper method.

frobel: We have 2 [PageContexts](https://github.com/shopware/acceptance-test-suite/blob/trunk/src/fixtures/PageContexts.ts) which we are using for our PageObjects. One is AdminPage and one is StorefrontPage. That 2 we inject per PageObject call ([see here](https://github.com/shopware/acceptance-test-suite/blob/d989aaa899092b1c136d8b29f0f4a2100b84ac34/src/page-objects/AdministrationPages.ts#L112))

So that means, our PageObject called e.g. AdminDetailPage are just "Placeholder" for the AdminPage page object etc.

Therefore, you could create 2 functions. One for admin pages, with AdminPage fixture and one for storefront pages with StorefrontPage fixture. 

like:  

```
HideAdminElementsForScreenshot: [
        async ({ AdminPage }, use) => {

const fn = async (selectors: string[]) => {
            if (!selectors.length) return;

            const css = selectors
            .map(selector => `${selector} { display: none !important; }`)
            .join('\n');

            await AdminPage.page.addStyleTag({ content: css });
        };

        await use(fn);
        },
        { scope: 'worker' },

```

That's just an idea.

ocavli: Done ✅ Could find the best solution so far and implemented the one you advised.

kstala (approved):

src/fixtures/HelperFixtures.ts
ocavli: I actually tried implementing it by injecting the method into the AdminPage and StorefrontPage objects. It definitely makes the usage more flexible, as you mentioned.

However, this approach caused some issues due to the scope mismatch between test and worker fixtures — especially in CI when using --repeat-each. You can see the related error here:[ [link to the error]](https://github.com/shopware/acceptance-test-suite/actions/runs/16049294930/job/45287897412?pr=413)

If it is ok with you, I’d suggest continuing with the current version for now.

frobel (approved):

PR #435, opened 2025-08-13
fix: change password auth to integration

GitEvil:
Updates the SaasInstanceSetup to accommodate the upcoming integration of SSO login.


MalteJanz:
We can close this as it's not needed anymore

PR #454, opened 2025-08-28
fix: disable services locator tests in SaaS

frobel:
- Disable services locator tests in SaaS because of possible race condition (one cloud instance for all tests) -> services are changing the whole instance behaviour, it is risky if that test fails during the execution

### DIFF
--- a/tests/ShopwareServices.spec.ts
+++ b/tests/ShopwareServices.spec.ts
@@ -1,0 +1,41 @@
+import { test, expect } from '../src';
+import { satisfies } from 'compare-versions';
+
+test('Shopware Services', async ({
+    InstanceMeta,
+    ShopAdmin,
+    AdminDashboard,
+    AdminShopwareServices,
+}) => {
+
+// eslint-disable-next-line playwright/no-conditional-in-test
+    if (satisfies(InstanceMeta.version, '>=6.7.1') && !InstanceMeta.isSaaS) {
+        await ShopAdmin.expects(AdminDashboard.shopwareServicesAdvertisementBanner).toBeVisible();
+        await ShopAdmin.expects(AdminDashboard.shopwareServicesAdvertisementBanner).toContainText('Introducing Shopware Services');
+        await ShopAdmin.expects(AdminDashboard.shopwareServicesExploreNowButton).toBeVisible();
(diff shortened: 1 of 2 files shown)


ocavli (approved):

sguder (approved):

tests/ShopwareServices.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+        const disableResponse = await disableResponsePromise;
+        expect(disableResponse.ok()).toBeTruthy();
+        // enable the services again for further tests
+        await ShopAdmin.expects(AdminShopwareServices.activateServicesButton).toBeVisible({ timeout: 15000 });
+        const enableResponsePromise = AdminShopwareServices.page.waitForResponse(`${ process.env['APP_URL'] }api/services/enable`);
+        await AdminShopwareServices.activateServicesButton.click();
+        const enableResponse = await enableResponsePromise;
+        expect(enableResponse.ok()).toBeTruthy();
+        await AdminShopwareServices.page.reload();
+        await ShopAdmin.expects(AdminShopwareServices.deactivateServicesButton).toBeVisible({ timeout: 15000 });
```
yusufttur, line 39: Why don't you use this task here?
src/tasks/shop-admin/ShopwareServices/DeactivateShopwareServices.ts

yusufttur, line 39: Is it possible to activate so fast? I guess you mentioned once that it uses a scheduled task or something similar and takes a bit of time to activate services?

yusufttur (changes requested):

tests/ShopwareServices.spec.ts
frobel: because I test here the locators specifically and in the task I couldnt do it. So the purpose is different.

frobel: I'm testing here the Locators and not the specific service. And currently, the services are not available at all on our pipelines, which are running on localhost. They are just visible and available on a publicly available address. We are on it that this behaviour would be possible too with localhost

yusufttur (approved):

ravenokko (approved):

PR #458, opened 2025-09-01
fix: empty state selector

FlorianKe:
Due to the switch from `sw-empty-state` to `mt-empty-state`, I need to update the empty-state selector.

relates shopware/shopware#9515


src/page-objects/administration/RuleDetail.ts
```diff
@@ -36,19 +36,19 @@ export class RuleDetail extends RuleCreate implements PageObject {
         this.shippingMethodAvailabilityRulesCard = page.locator('.sw-settings-rule-detail-assignments__card-shipping_method_availability_rule');
         this.shippingMethodAvailabilityRulesCardLink = this.shippingMethodAvailabilityRulesCard.getByRole('link');
         this.shippingMethodAvailabilityRulesCardTable = page.locator('.sw-settings-rule-detail-assignments__entity-listing-shipping_method_availability_rule');
-        this.shippingMethodAvailabilityRulesCardEmptyState = this.shippingMethodAvailabilityRulesCard.getByRole('alert');
+        this.shippingMethodAvailabilityRulesCardEmptyState = this.shippingMethodAvailabilityRulesCard.locator('.mt-empty-state__headline');
```
frobel: We have to make sure that the locator is backwards compatible (working with latest 6.6.x version and 6.7.x versions)

sguder (approved):

PR #461, opened 2025-09-04
feat: add Media page objects

vanpham-sw:


src/page-objects/administration/Media.ts
```diff
@@ -0,0 +1,72 @@
...
+  public readonly uploadFileButton: Locator;
+  public readonly addNewFolderButton: Locator;
+  public readonly mediaGridItems: Locator;
+  public readonly mediaItemCheckbox: Locator;
+  public readonly mediaItemPreview: Locator;
+  public readonly mediaItemName: Locator;
+  public readonly mediaItemContextMenu: Locator;
+  public readonly searchInput: Locator;
+
+
```
vanpham-sw:
```suggestion
```

src/page-objects/administration/Media.ts
```diff
@@ -0,0 +1,71 @@
...
+    return `#/sw/media/index`;
+  }
+  
+  /**
+   * Gets a specific media item by its name
+   * @param name The filename to look for
+   * @returns Locator for the specific media item
+   */
+  getMediaItemByName(name: string): Locator {
+    return this.page.locator(`.sw-media-base-item__name`, { hasText: name }).first();
```
frobel: Why first? Should we not be sure we have a unique name here?

src/page-objects/administration/Media.ts
```diff
@@ -0,0 +1,71 @@
...
+    return this.mediaGridItems.filter({ hasText: name }).locator('.sw-context-button');
+  }
+  
+  /**
+   * Gets a specific context menu item by its text
+   * @param itemText The text of the context menu item (e.g., 'Rename', 'Delete', etc.)
+   * @returns Locator for the specific context menu item
+   */
+  getContextMenuItem(itemText: string): Locator {
+    return this.page.locator('.sw-context-menu-item', {
```
frobel, line 69: Should we use here the bind element: `mediaItemContextMenu` so that we just look at that context?

vanpham-sw: @frobel  The context menu popup is located in another tree so I couldn't bind it to the media item:

image

frobel: i see...again a popover menu :)

frobel (approved):

ravenokko (approved):

PR #476, opened 2025-09-24
fix: additional element for account pages

### LINKED ISSUE #8655: Visual test // Storefront: Account
### User story

Create test structure and take screenshots at the following places.

### Acceptance criteria

Screenshots of these Account pages:

- Overview
- Profile
- Adresses
- Orders
- Login

ocavli:


src/page-objects/storefront/Account.ts
```diff
@@ -17,6 +17,14 @@ export class Account implements PageObject {
     public readonly page: Page;
     public readonly instanceMeta: HelperFixtureTypes['InstanceMeta'];
 
+    //Account menu
```
vanpham-sw: Can we implement something common for the Account menu ,without duplicating across pages?

src/page-objects/storefront/Account.ts
```diff
@@ -35,6 +43,14 @@ export class Account implements PageObject {
         }
         this.cannotDeliverToCountryAlert = page.getByText('We can not deliver to the country that is stored in your delivery address.');
         this.shippingToAddressNotPossibleAlert = page.getByText('Shipping to the selected shipping address is currently not possible.');
+
+        //Account menu
+        this.overviewLink = page.getByRole('link', { name: 'Overview' });
+        this.yourProfileLink = page.getByRole('link', { name: 'Your profile' });
+        this.addressesLink = page.getByRole('link', { name: 'Addresses' });
+        this.ordersLink = page.getByRole('link', { name: 'Orders' });
+        this.subscriptionsLink = page.getByRole('link', { name: 'Subscriptions' });
```
frobel: Just commercial related I guess or?

src/page-objects/storefront/Account.ts
frobel: Good point @vanpham-sw . We have to introduce here a new logic so my idea after some research:

- Create a new Component (class) called e.g. `account-nav.component.ts`
- Put the navigation items into it:
```
export class AccountNav {

    public readonly overviewLink: Locator;
    public readonly yourProfileLink: Locator;
    public readonly addressesLink: Locator;
    public readonly ordersLink: Locator;
    public readonly logoutLink: Locator;

  constructor(private readonly page: Page) {
        this.overviewLink = page.getByRole('link', { name: 'Overview' });
        this.yourProfileLink = page.getByRole('link', { name: 'Your profile' });
        this.addressesLink = page.getByRole('link', { name: 'Addresses' });
        this.ordersLink = page.getByRole('link', { name: 'Orders' });
this.logoutLink = page.getByRole('link', { name: 'Log out' });
}

```

- Next step is introducing a BaseAccountPage which contains the navigation component
```
// base-account.ts

export abstract class BaseAccountPage {
  readonly page: Page;
  readonly nav: AccountNav;
  readonly footer: AccountFooterNav;

  constructor(page: Page) {
    this.page = page;
    this.nav = new AccountNav(page);         
  }
}

```
In the last step the actual page objects like AccountProfil, AccountOrder etc extends it with
```
// account-orders.page.ts

export class AccountOrdersPage extends BaseAccountPage {
  constructor(page: Page) { 
super(page); 
}
}

```

So why that approach and not just extending an AccountNav Page object:
Extending AccountNav makes each page “is-a nav”, which is a leaky abstraction. Pages should have a nav, not be one. It also paints us into a corner the moment we add a second shared component (e.g., Footer, Sidebar etc)—TypeScript has no multiple inheritance option. So we could then simply add a new Instance in the BaseAccountPage with the keyword `new` in the constructor (`new FooterNav()` )

frobel (changes requested):

src/page-objects/storefront/Account.ts
ocavli: Thanks for the hint 👍 Updated the pages and added the components folder in storefront. If that is ok for you? And the test works fine with that solution.

src/page-objects/storefront/Account.ts
ocavli: Yes, good catch, removed this one.

vanpham-sw (approved):

src/page-objects/storefront/AccountAddresses.ts
```diff
@@ -16,11 +16,10 @@ export class AccountAddresses implements PageObject {
     public readonly addressDropdownButtons: Locator | undefined;
     public readonly availableAddressesUseAsBillingAddress: Locator | undefined;
     public readonly availableAddressesUseAsShippingAddress: Locator | undefined;
-    public readonly page: Page;
     public readonly instanceMeta: HelperFixtureTypes['InstanceMeta'];
 
-    constructor(page: Page, instanceMeta: HelperFixtureTypes['InstanceMeta']) {
-        this.page = page;
+    constructor(page: Page,instanceMeta: HelperFixtureTypes['InstanceMeta']) {
```
frobel:
```suggestion
    constructor(page: Page, instanceMeta: HelperFixtureTypes['InstanceMeta']) {
```

frobel (changes requested):

src/page-objects/storefront/components/AccountNavigation.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export class AccountNavigation {
+  public readonly overviewLink: Locator;
+  public readonly yourProfileLink: Locator;
+  public readonly addressesLink: Locator;
+  public readonly ordersLink: Locator;
+  public readonly logoutLink: Locator;
+
+  constructor(private readonly page: Page) {
+    this.overviewLink = page.getByRole('link', { name: 'Overview' });
```
hienha-sw: Please consider to use stable locators before using translation key
- data-testid (best practice)
- stable attributes (eg. name="email", id="workflow-name", type="password")
- ARIA roles with scoping
- visible text with translation key (last resort) refer [this PR](https://github.com/shopware/acceptance-test-suite/pull/478)

src/page-objects/storefront/components/AccountNavigation.ts
```diff
@@ -0,0 +1,17 @@
...
+    this.yourProfileLink = page.getByRole('link', { name: 'Your profile' });
+    this.addressesLink = page.getByRole('link', { name: 'Addresses' });
+    this.ordersLink = page.getByRole('link', { name: 'Orders' });
+    this.logoutLink = page.getByRole('link', { name: 'Log out' });
+  }
+}
```
hienha-sw: Should enforce line endings

hienha-sw (changes requested):

hienha-sw (approved):

src/page-objects/storefront/components/AccountNavigation.ts
ocavli: thx for the hint 👍

frobel (approved):

PR #491, opened 2025-10-06
refactor: update current test suite to include keyboard functionality and a11y checks

ravenokko:
Integration of keyboard functionality and a11y checks as defined in https://github.com/shopware/acceptance-test-suite/pull/410


hienha-sw (approved):

frobel (approved):

yusufttur (approved):

src/locales/en/storefront/home.json
```diff
@@ -3,16 +3,20 @@
...
     },
     "filters": {
+        "filterPanel": "Filter products",
+        "labelPrefix": "Filter by ",
         "manufacturer": "Manufacturer",
         "price": "Price",
         "resetAll": "Reset all",
-        "freeShipping": "Free shipping"
+        "freeShipping": "Add filter: Free shipping",
+        "rating": "minimum rating"
```
frobel, line 20: Everything in small letters?

src/page-objects/storefront/Wishlist.ts
```diff
@@ -20,7 +21,7 @@ export class Wishlist extends Home implements PageObject {
     async getListingItemByProductName(productListingName: string): Promise<Record<string, Locator>> {
         const baseItems = await super.getListingItemByProductName(productListingName);
         const listingItem = this.page.getByRole('listitem').filter({ has: this.page.getByText(productListingName) });
-        const removeFromWishlistButton = listingItem.locator('.icon-wishlist-remove');
+        const removeFromWishlistButton = listingItem.getByTitle(translate('storefront:wishlist:removeProduct'));//listingItem.getByRole('button', { name: `Remove ${productListingName} from wishlist` });
```
frobel: why uncomments here?

src/tasks/shop-customer/Account/LoginViaReviewsTab.ts
```diff
@@ -8,16 +8,18 @@ export const LoginViaReviewsTab = base.extend<{ LoginViaReviewsTab: Task }, Fixt
         const task = (product: Product, customCustomer?: Customer) => {
             return async function LoginViaReviewsTab() {
 
+                const loginResponse = await StorefrontProductDetail.page.request.post('account/login');
```
frobel, line 11: why that step is necessary?

src/tasks/shop-customer/Account/Register.ts
```diff
@@ -37,29 +37,33 @@ export const Register = base.extend<{ Register: Task }, FixtureTypes>({
 
                 registeredEmail = registrationData.email;
 
+                await ShopCustomer.presses(StorefrontAccountLogin.salutationSelect);
                 await StorefrontAccountLogin.salutationSelect.selectOption(registrationData.salutation);
```
frobel, line 41: Both are same? Or why we need the presses call before?

src/tasks/shop-customer/Product/SelectProductFilterOption.ts
```diff
@@ -4,12 +4,14 @@ import type { Task } from '../../../types/Task';
...
 export const SelectProductFilterOption = base.extend<{ SelectProductFilterOption: Task }, FixtureTypes>({
-    SelectProductFilterOption: async ({ StorefrontHome }, use) => {
+    SelectProductFilterOption: async ({ ShopCustomer, StorefrontHome }, use) => {
         const task = (filterButton: Locator, optionName: string) => {
             return async function SelectProductFilterOption() {
-                await filterButton.click();
+                await ShopCustomer.presses(filterButton);
                 const optionLocator = await StorefrontHome.getFilterItemByFilterName(optionName);
-                await optionLocator.click();
+                ShopCustomer.expects(optionLocator).not.toBeChecked();
```
frobel: `await` is not missing here or?

frobel (changes requested):

src/tasks/shop-customer/Account/Register.ts
ravenokko: Not quite. `presses()` will evaluate if, when focused, it has a visible focus indicator and yes, will open the dropdown (but that doesn't seem to cause any problems). However, because `<select>` is a native element Playwright has to do some voodoo because the open dropdown is rendered in the browser, not the DOM (this is why they have a separate function and also why you can't just use `click()` to pick an option). So to actually check for a visible focus -and- select an option we need both.

src/locales/en/storefront/home.json
ravenokko: The product rating filter has the aria-label `Filter by minimum rating` which is different from the visible text `Rating min.`.  I use that label since all filters (except Free Shipping) have the aria-label `Filter by {name}` which is a pattern I could then enforce in line 150 of the Home.ts page object

```
async getFilterButtonByFilterName(filterName: string): Promise<Locator> {
        return this.filterMultiSelect.getByRole('button', 
        { name: `${translate('storefront:home:filters.labelPrefix').concat(filterName)}` });
```

src/page-objects/storefront/Wishlist.ts
ravenokko: for the very important reason that I forgot to remove them

src/tasks/shop-customer/Account/LoginViaReviewsTab.ts
ravenokko: This was originally added to the [Review test to fix it from failing in Saas](https://github.com/shopware/shopware/pull/11127). I've been meaning to move it to this task so that it won't cause problems for any future tests we might need

src/tasks/shop-customer/Product/SelectProductFilterOption.ts
ravenokko: fixed! thank you!

frobel (approved):

PR #508, opened 2025-10-20
test: add test for custom matchers

ravenokko:
ensure `toHaveVisibleFocus()` can successfully detect outline/box-shadow focus styles of existing elements

### DIFF
--- a/src/locales/index.ts
+++ b/src/locales/index.ts
@@ -29,7 +29,7 @@ import storefrontProduct from './en/storefront/product.json' with { type: 'json'
 import storefrontNavigation from './en/storefront/navigation.json' with { type: 'json' };
 import storefrontContact from './en/storefront/contact.json' with { type: 'json' };
 import storefrontConsent from './en/storefront/consent.json' with { type: 'json' };
-import StorefrontHeader from './en/storefront/header.json' with { type: 'json' };
+import storefrontHeader from './en/storefront/header.json' with { type: 'json' };
 import storefrontHome from './en/storefront/home.json' with { type: 'json' };
 import storefrontLogin from './en/storefront/login.json' with { type: 'json' };
 import storefrontOrder from './en/storefront/order.json' with { type: 'json' };
@@ -113,6 +113,7 @@ export const BUNDLED_RESOURCES = {
         'storefront/navigation': storefrontNavigation,
         'storefront/contact': storefrontContact,
         'storefront/consent': storefrontConsent,
+        'storefront/header': storefrontHeader,
         'storefront/home': storefrontHome,
         'storefront/login': storefrontLogin,
         'storefront/order': storefrontOrder,
@@ -198,7 +199,7 @@ export const baseNamespaces = {
         navigation: storefrontNavigation,
         contact: storefrontContact,
         consent: storefrontConsent,
-        header: StorefrontHeader,
+        header: storefrontHeader,
         home: storefrontHome,
         login: storefrontLogin,
         order: storefrontOrder,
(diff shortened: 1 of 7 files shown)


src/page-objects/storefront/Header.ts
```diff
@@ -19,7 +23,11 @@ export class Header implements PageObject {
         this.languagesDropdown = page.locator('.top-bar-language').filter({ has: page.getByRole('button') });
         this.languagesMenuOptions = page.locator('.top-bar-language').filter({ has: page.getByRole('list') });
         this.currenciesDropdown = page.locator('.top-bar-currency').filter({ has: page.getByRole('button') });
-        this.currenciesMenuOptions = page.locator('.top-bar-currency').filter({ has: page.getByRole('list') });        
+        this.currenciesMenuOptions = page.locator('.top-bar-currency').filter({ has: page.getByRole('list') });
+        this.searchInput = page.getByRole('combobox', { name: 'Search' });        
+
+        //skip links
+        this.skipToMainContentLink = page.getByRole('link', { name: 'Skip to main content' });
```
hienha-sw: Please update with translation key

frobel (approved):

src/page-objects/storefront/Header.ts
ravenokko: Done for `skipToMainContentLink`. The `searchInput`, however, doesn't use a snippet - the placeholder text/aria label changes, but the `name` attribute is always 'search' no matter what language is selected. So I left it as is.

hienha-sw: Do you mean this Search icon? I noticed it’s not labeled as `search` when the language is set to German.
image
Even if the label happens to be the same in both English and German, it might differ in other languages. I’m not completely sure, but we should make it consistent to avoid confusion and potential issues when adding more languages later on.

src/locales/index.ts
```diff
@@ -29,6 +29,7 @@ import storefrontProduct from './en/storefront/product.json' with { type: 'json'
 import storefrontNavigation from './en/storefront/navigation.json' with { type: 'json' };
 import storefrontContact from './en/storefront/contact.json' with { type: 'json' };
 import storefrontConsent from './en/storefront/consent.json' with { type: 'json' };
+import StorefrontHeader from './en/storefront/header.json' with { type: 'json' };
```
hienha-sw: StorefrontHeader -> storefrontHeader

src/locales/index.ts
```diff
@@ -149,6 +151,7 @@ export const BUNDLED_RESOURCES = {
         'storefront/checkout': deStorefrontCheckout,
         'storefront/consent': deStorefrontConsent,
         'storefront/contact': deStorefrontContact,
+        'storefront/header': deStorefrontHeader,
```
hienha-sw, line 155: Please add it to en: {} too — I tried to keep the updates minimal.

hienha-sw:
The tests are failing — could you take a look?

src/page-objects/storefront/Header.ts
ravenokko: No it's this:

Screenshot 2025-10-22 at 14 31 29

There's an ARIA label that updates based on language (obviously) but there's an explicit attribute `name="search"` which never changes

Screenshot 2025-10-22 at 14 51 59

I wasn't sure if we wanted to mix the ARIA label in with snippets. But it seems like [we do that in platform](https://github.com/shopware/shopware/blob/c116bbc5bdd1e2659ca60312e8e7e2f615abc068/src/Storefront/Resources/snippet/storefront.en.json#L89).  Technically, it doesn't even need a name because it's the only input box in the header, but I'd prefer if it had one for error handling. I'll add the ARIA label as a snippet then if there are no objections 🙂

ravenokko: updated header.json to include aria labels and redefined locator to use getByLabel()

hienha-sw (approved):

PR #515, opened 2025-11-05
fix: update locale_mappings:  languageCode of en-US locale to en-GB

hienha-sw:
Shopware do not support languageCode `en-US`, so update languageCode of 'en-US' locale to `en-GB`

### DIFF
--- a/src/services/ShopwareDataHelpers.ts
+++ b/src/services/ShopwareDataHelpers.ts
@@ -11,7 +11,7 @@ interface LocaleMapping {
 }
 
 const LOCALE_MAPPINGS: Readonly<Record<string, LocaleMapping>> = {
-    'en-US': { countryCode: 'US', currencyCode: 'USD', currencySymbol: '$', languageCode: 'en-US' },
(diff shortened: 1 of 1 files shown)


src/services/ShopwareDataHelpers.ts
```diff
@@ -11,7 +11,7 @@ interface LocaleMapping {
 }
 
 const LOCALE_MAPPINGS: Readonly<Record<string, LocaleMapping>> = {
-    'en-US': { countryCode: 'US', currencyCode: 'USD', currencySymbol: '$', languageCode: 'en-US' },
+    'en-US': { countryCode: 'US', currencyCode: 'USD', currencySymbol: '$', languageCode: 'en-GB' },
```
frobel, line 14: @ravenokko how that is managed in the LanguagePack Plugin 🤔

frobel (approved):

src/services/ShopwareDataHelpers.ts
ravenokko: It is `en-US` in the language pack, the locale mapping definition was correct. We've also integrated it so all languages can be installed via command line in the core platform - but `en-GB` and `de-DE` are still the system defaults unless otherwise specified in the installer.

PR #521, opened 2025-11-07
feat: update selectors for dynamic field rendering with meteor components

jleifeld:
Update tests for https://github.com/shopware/shopware/pull/13185

### DIFF
--- a/src/page-objects/administration/RuleDetail.ts
+++ b/src/page-objects/administration/RuleDetail.ts
@@ -113,7 +113,11 @@ export class RuleDetail extends RuleCreate implements PageObject {
         this.conditionCartLineItemTaxationValue = page.locator('.sw-condition__condition-type-cartLineItemTaxation').locator('.sw-select-selection-list__item');
         this.conditionTimeRangeValueFirst = page.locator('.sw-condition-time-range').getByRole('textbox').first()
         this.conditionTimeRangeValueSecond = page.locator('.sw-condition-time-range').getByRole('textbox').last()
-        this.conditionOrderCreatedByAdminValue = page.locator('.sw-condition__condition-type-orderCreatedByAdmin').locator('.sw-select__selection');
+        if (satisfies(instanceMeta.version, '<6.7.4.1')) {
+            this.conditionOrderCreatedByAdminValue = page.locator('.sw-condition__condition-type-orderCreatedByAdmin').locator('.sw-select__selection');        
+        } else {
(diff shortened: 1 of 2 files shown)


src/page-objects/administration/RuleDetail.ts
```diff
@@ -113,7 +113,7 @@ export class RuleDetail extends RuleCreate implements PageObject {
         this.conditionCartLineItemTaxationValue = page.locator('.sw-condition__condition-type-cartLineItemTaxation').locator('.sw-select-selection-list__item');
         this.conditionTimeRangeValueFirst = page.locator('.sw-condition-time-range').getByRole('textbox').first()
         this.conditionTimeRangeValueSecond = page.locator('.sw-condition-time-range').getByRole('textbox').last()
-        this.conditionOrderCreatedByAdminValue = page.locator('.sw-condition__condition-type-orderCreatedByAdmin').locator('.sw-select__selection');
+        this.conditionOrderCreatedByAdminValue = page.locator('.sw-condition__condition-type-orderCreatedByAdmin').locator('.mt-select-selection-list__input');
```
frobel: We have to introduce here some `IF` checks because we introduce it in a specific version. 

So our tests will fail with 6.6. and 6.7. first but will be green with 6.7. latest.

src/page-objects/administration/modules/CustomFieldCard.ts
```diff
@@ -20,7 +20,7 @@ export async function getCustomFieldCardLocators(page: Page, customFieldSetName:
         customFieldCard = page.locator('.mt-card').filter({ hasText: translate('administration:customField:general.customFields') });
     }
     const customFieldSetTab = customFieldCard.getByText(customFieldSetName);
-    const customFieldLabel = customFieldCard.locator('.sw-custom-field-set-renderer').locator('.sw-field__label').getByText(customFieldName);
+    const customFieldLabel = customFieldCard.locator('.sw-custom-field-set-renderer').locator('.mt-field__label').getByText(customFieldName);
```
frobel: We have to introduce here some `IF` checks because we introduce it in a specific version. 

So our tests will fail with 6.6. and 6.7. first but will be green with 6.7. latest.

frobel:
```
if (satisfies(instanceMeta.version, '<6.7.4???')) {
... .sw-field__label
} else {
.mt-field__label
}


```

frobel (changes requested):

frobel (approved):

PR #528, opened 2025-11-13
fix: review filter rating locator

yusufttur:
`getByLabel` doesn't work with the latest changes.
https://github.com/shopware/shopware/pull/13493

### DIFF
--- a/src/page-objects/storefront/Home.ts
+++ b/src/page-objects/storefront/Home.ts
@@ -139,7 +139,7 @@ export class Home implements PageObject {
     }
 
     async getRatingItemLocatorByRating(rating: number): Promise<Locator> {
-        return this.productRatingList.getByLabel(`min. ${rating}/5`);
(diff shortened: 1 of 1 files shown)


src/page-objects/storefront/Home.ts
```diff
@@ -139,7 +139,7 @@ export class Home implements PageObject {
     }
 
     async getRatingItemLocatorByRating(rating: number): Promise<Locator> {
-        return this.productRatingList.getByLabel(`min. ${rating}/5`);
+        return this.productRatingList.getByText(`min. ${rating}/5`);
```
frobel, line 142: Is getByText backwards compatible with 6.6.x?

yusufttur: It should since it uses the same text. 
Screenshot 2025-11-13 at 13 25 17

frobel (approved):

PR #544, opened 2025-11-26
fix: presses and navigation stability

ravenokko:
The `Go to Checkout ` link on the off-canvas was flaky because Playwright would try to focus the link sometimes before the canvas was finished loading/animating. 

Additionally, I noticed some potential for flakiness with `Submit Order` and `Go To Checkout`. Pressing (or clicking) these links create a redirect which Playwright has no knowledge of, so I added the automatic waits in as is [best practice](https://playwright.dev/docs/navigations).

### DIFF
--- a/src/tasks/shop-customer/Cart/ProceedFromCartToCheckout.ts.ts
+++ b/src/tasks/shop-customer/Cart/ProceedFromCartToCheckout.ts.ts
@@ -7,7 +7,7 @@ export const ProceedFromCartToCheckout = base.extend<{ ProceedFromCartToCheckout
         ShopCustomer,
         StorefrontCheckoutCart,
         StorefrontCheckoutConfirm,
-        StorefrontPage
+        StorefrontPage,
     }, use) => {
         const task = () => {
             return async function ProceedFromCartToCheckout() {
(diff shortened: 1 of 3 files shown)


vanpham-sw (approved):

src/tasks/shop-customer/Cart/ProceedFromCartToCheckout.ts.ts
```diff
@@ -7,10 +7,12 @@ export const ProceedFromCartToCheckout = base.extend<{ ProceedFromCartToCheckout
         ShopCustomer,
         StorefrontCheckoutCart,
         StorefrontCheckoutConfirm,
+        StorefrontPage,
     }, use) => {
         const task = () => {
             return async function ProceedFromCartToCheckout() {
-                await StorefrontCheckoutCart.goToCheckoutButton.click();
+                await ShopCustomer.presses(StorefrontCheckoutCart.goToCheckoutButton);
+                await StorefrontPage.waitForURL('**/checkout/confirm', { waitUntil: 'commit' });
```
frobel: 1. Instead of a new Fixture called StorefrontPage you can also use `StorefrontCheckoutCart.page.waitForURL()`

2. Why you waitUntil "commit" ? And not after load 🤔 ? reference see [here](https://playwright.dev/docs/api/class-page#page-wait-for-url)

ravenokko: 1. Thanks, I was not 100% sure I could call `waitForURL()` on the page object I wanted to load - I'll change it back.
2. Because this is the most performant - we only need to prevent Playwright from immediately checking for the headline on the product page after pressing the button. 9 out of 10 times this task already passes without the `waitForURL()` because of the built in auto-waiting for the headline assertion.   This way helps us both enforce quick load times for the storefront, while ensuring the task is more stable.

frobel (approved):

PR #556, opened 2025-12-18
feat: make closing the off canvas cart a task

sguder:


src/tasks/shop-customer/Cart/CloseTheOffCanvasCart.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const CloseTheOffCanvasCart = base.extend<{ CloseTheOffCanvasCart: Task }, FixtureTypes>({
+    CloseTheOffCanvasCart: async ({
+        ShopCustomer,
+        StorefrontOffCanvasCart,
+    }, use) => {
+        const task = () => {
+            return async function CloseTheOffCanvasCart() {
+                await StorefrontOffCanvasCart.continueShoppingButton.scrollIntoViewIfNeeded();
+                await ShopCustomer.expects(StorefrontOffCanvasCart.continueShoppingButton).toBeVisible();
```
ravenokko: Can remove this line because it's redundant. The `a11y_checks()` in `presses()` includes and asserts `focus()` - this inherently checks if an element is visible (or at least, no 'hidden' attribute) and active/enabled. The only thing it can't check is stability which is why we need `scrollIntoViewIfNeeded()`.

ravenokko (changes requested):

src/tasks/shop-customer/Cart/CloseTheOffCanvasCart.ts
sguder: removed

ravenokko (approved):

PR #559, opened 2025-12-30
fix: use StoreApiContext for isThemeCompiled

mkraeml:
We should use the StoreApiContext for checking the `all.css`. The AdminApiContext uses the Admin Bearer Token and as we now have a better error reporting on the saas CDN,  the isThemeCompiled check will fail on saas.


mkraeml:
The failing `test-saas-in-german` test is not caused by my change

frobel (approved):

PR #562, opened 2026-01-06
fix: refactor Actor.selectsRadioButton function

### LINKED ISSUE #558: Refactor "selectsRadioButton" function in Actor.ts
The selectsRadioButton function right now has two issues:

Image

1) The "===" check is too strict, often labels contain more than just an identifying string and can even contain further elements

2) The search for every label in the radioGroup, detemining how many steps are needed to reach that label and then just press the arrow key that many times does work only, if each radio button has just one label element. If there are multiple label elements per radio button (like on the product detail page, when you have subscriptions activated), it doesn't work anymore.

Solutions:
1) Change line 78 to the following should do the trick:
`const desiredOptionIndex = options.findIndex(opt => { return opt.label.includes(inputLabel) });`

2) Maybe instead of finding the label index to determine how many times we need to press the arrow key, we could just press the arrow key in the loop and check every iteration, if the radio button currently focused has the label we want to have. This way it doesn't matter how many additional labels are existing on the lower levels

sguder:
Closes #558 (see for description)

### DIFF
--- a/src/services/Actor.ts
+++ b/src/services/Actor.ts
@@ -14,7 +14,7 @@ export class Actor {
 
     expects = expect;
 
-    async a11y_checks(locator: Locator){
+    async a11y_checks(locator: Locator) {
         await locator.focus();
         await expect(locator).toBeFocused();        
         await expect(locator).toHaveVisibleFocus(); 
--- a/src/services/Actor.ts
+++ b/src/services/Actor.ts
@@ -67,7 +67,7 @@ export class Actor {
                 return;
             }
         
-            const checkedRadio = radioGroup.getByRole('radio', { checked: true });
+            let checkedRadio = radioGroup.getByRole('radio', { checked: true });
         
             if (!(await checkedRadio.count())) {
                 throw new Error('No radio button is selected by default.');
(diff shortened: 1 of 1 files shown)


src/services/Actor.ts
```diff
@@ -56,44 +56,39 @@ export class Actor {
...
+        
+            const maxIterations = await radioGroup.getByRole('radio').count();
+            let iterations = 0;
+        
+            while (!(await desiredOption.isChecked())) {
+                if (iterations >= maxIterations) {
+                    throw new Error(`Could not reach radio button "${inputLabel}" via keyboard navigation.`);
                 }
+        
+                await this.page.keyboard.press('ArrowDown');
```
ravenokko: `presses()` would be better so that `a11y-checks()` is then run on every radio button as it cycles through.

yusufttur (approved):

ravenokko (changes requested):

src/services/Actor.ts
```diff
@@ -56,44 +56,39 @@ export class Actor {
...
+                return;
+            }
+        
+            const checkedRadio = radioGroup.getByRole('radio', { checked: true });
+        
+            if (!(await checkedRadio.count())) {
+                throw new Error('No radio button is selected by default.');
+            }
+        
+            await checkedRadio.first().focus();
```
ravenokko: what does this do, exactly?

sguder: Sets focus on the element and scrolls it into view if needed, so we have a starting point for keyboard navigation

ravenokko (approved):

src/services/Actor.ts
sguder: done

PR #565, opened 2026-01-07
fix: remove unnecessary scroll function calls

sguder:
These are no longer necessary with https://github.com/shopware/acceptance-test-suite/pull/563

### DIFF
--- a/src/tasks/shop-customer/Cart/CloseTheOffCanvasCart.ts
+++ b/src/tasks/shop-customer/Cart/CloseTheOffCanvasCart.ts
@@ -10,6 +10,7 @@ export const CloseTheOffCanvasCart = base.extend<{ CloseTheOffCanvasCart: Task }
         const task = () => {
             return async function CloseTheOffCanvasCart() {
                 await ShopCustomer.presses(StorefrontOffCanvasCart.continueShoppingButton);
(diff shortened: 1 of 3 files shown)


src/tasks/shop-customer/Cart/CloseTheOffCanvasCart.ts
```diff
@@ -9,7 +9,6 @@ export const CloseTheOffCanvasCart = base.extend<{ CloseTheOffCanvasCart: Task }
     }, use) => {
         const task = () => {
             return async function CloseTheOffCanvasCart() {
-                await StorefrontOffCanvasCart.continueShoppingButton.scrollIntoViewIfNeeded();
                 await ShopCustomer.presses(StorefrontOffCanvasCart.continueShoppingButton);
```
ravenokko, line 12:
```suggestion
                await ShopCustomer.presses(StorefrontOffCanvasCart.continueShoppingButton);
                await ShopCustomer.expects(StorefrontOffCanvasCart.continueShoppingButton).not.toBeVisible();
```

ravenokko: since pressing the button closes the menu it'd be better for us to assert the menu did, in fact, close

ravenokko (approved):

PR #568, opened 2026-02-02
fix: Add exact match check to Admin Log in button

ravenokko:
We have since added a `Log in with Shopware` button for SSO (which used to be a link). Without `exact:true` the `Log in `button resolves to 2 elements.

Additionally, the `StorefrontPageObjects` test was flaky in Saas - sometimes the category page wouldn't show the product that had been added. By appending the current date/time to the category URL we can be certain to be viewing the correct page state after the cache has been cleared.

### DIFF
--- a/tests/PageObjects/StorefrontPageObjects.spec.ts
+++ b/tests/PageObjects/StorefrontPageObjects.spec.ts
@@ -22,7 +22,6 @@ test('Storefront page objects', async ({
     TestDataService,
     InstanceMeta,
     CheckVisibilityInHome,
-    StorefrontHome,
 }) => {
 
     const category = await TestDataService.createCategory();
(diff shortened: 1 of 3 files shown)


tests/PageObjects/StorefrontPageObjects.spec.ts
```diff
@@ -22,17 +22,15 @@ test('Storefront page objects', async ({
...
     CheckVisibilityInHome,
-    StorefrontHome,
 }) => {
 
-    const product = await TestDataService.createBasicProduct();
     const category = await TestDataService.createCategory();
+    const product = await TestDataService.createBasicProduct();
     await TestDataService.assignProductCategory(product.id, category.id);
 
-    await ShopCustomer.goesTo(StorefrontHome.url())
```
ravenokko, line 32: Removed as the ShopCustomer navigates to `StorefrontHome` as part of `CheckVisibilityInHome`

vanpham-sw (approved):

PR #570, opened 2026-02-05
refactor: migrate eslint to oxlint and improve linting

patzick:
Few cleanup things here:
- migrated eslint to oxlint with type-aware checking
- added GO version of typescript for typecheck command
- changed dependencies to strict versioning
- bumped versions in actions
- added dependabot to auto bump dependencies via PRs
- removed all eslint-disable comments

single lint warning left, because of skipped tests that should be dealt with or removed in future
image

Nerdy cool stats, lint before:
`npm run lint  2.15s user 0.25s system 147% cpu 1.630 total`
after with even more checks
`npm run lint  0.83s user 0.27s system 151% cpu 0.727 total`


patzick:
`eslint` check is "expected" here but i renamed it to just `lint` to make it independent from the tooling, the `lint` job itself passes and I'll adjust "expected" check after the merge

yusufttur (approved):

yusufttur:
@sguder, I think we can remove `Promotions Listing - Empty state` skipped test. Wdyt

PR #577, opened 2026-02-06
feat: add oxfmt formatter 

### LINKED ISSUE #571: Add formatter
follow up of #570 

there is prettier config in repo but format command is not specified and some files are not formatted with these rules. After merging mentioned PR I'll add oxfmt support and run format on files

patzick:
closes #571 

- added `oxfmt` as a formatter
- added `lefthook` with pre-commit hook to run lint format edited files 
- the rest of the changes are just formatter changes

this way files are always linted and formatted locally no matter the editor


yusufttur (approved):

src/data-fixtures/Category/Category.ts
```diff
@@ -1,29 +1,30 @@
...
  * @deprecated - Use TestDataService.createCategory() instead.
  */
 export const CategoryData = base.extend<FixtureTypes>({
     CategoryData: async ({ IdProvider, AdminApiContext, DefaultSalesChannel, ProductData }, use) => {
-
         const { id: categoryId, uuid: categoryUuid } = IdProvider.getIdPair();
         const categoryName = `Category-${categoryId}`;
 
-        const categoryResponse = await AdminApiContext.post('category?_response', {
+        const categoryResponse = await AdminApiContext.post("category?_response", {
```
frobel, line 13: Now we use " instead of ' ...is that shopware standard?

frobel (changes requested):

src/data-fixtures/Category/Category.ts
patzick: that's definetly broader standard than single quote and the closer we are to established opinionated formatters the better overall compatibility

If you have very stron opinion for single quote for this repo I'm open to set it this way, LMK!

frobel (approved):

frobel (approved):

PR #579, opened 2026-02-06
fix: migrate selector from sw-popover to mt-floating-ui

gweiermann:
# Don't merge until https://github.com/shopware/shopware/pull/14664 is merged
Due to migration in https://github.com/shopware/shopware/pull/14664, some cases of `<sw-popover />` were migrated to `<mt-floating-ui />`. This PR adjusts an acceptance test accordingly


src/page-objects/administration/FirstRunWizard.ts
```diff
@@ -85,7 +85,7 @@ export class FirstRunWizard implements PageObject {
         // Default values part
         this.defaultValuesHeader = page.locator(".sw-modal__title", { hasText: translate("administration:firstRunWizard:headers.defaultValues") });
         this.salesChannelSelectionMultiSelect = page.getByPlaceholder(translate("administration:firstRunWizard:placeholders.selectSalesChannels"));
-        this.salesChannelSelectionList = page.locator(".sw-popover__wrapper").getByRole("listitem");
+        this.salesChannelSelectionList = page.locator(".mt-floating-ui__content").getByRole("listitem");
```
frobel: The issue here is, that we have to make sure our acceptance tests are backwards compatible. We have to build in some version conditions. see example [here](https://github.com/shopware/acceptance-test-suite/blob/a9a4b0ee86acd977345a913a577cc96315e0b375/src/page-objects/administration/OrderListing.ts#L29): 

So maybe you already know in with version the new feature will be shipped? So you can use that version.

frobel (changes requested):

src/page-objects/administration/FirstRunWizard.ts
gweiermann: fixed it with a more generic selector that works for all versions

ravenokko (approved):

PR #591, opened 2026-02-24
feat: add additional locators on storefront address page

sguder:


src/page-objects/storefront/AccountAddresses.ts
```diff
@@ -42,4 +43,47 @@ export class AccountAddresses extends BaseAccount {
...
+
+        if (!this.availableAddresses) {
+            throw new Error("Available addresses are not defined.");
+        }
+
+        const addressContainer = this.availableAddresses.locator(".address-manager-select-address").filter({ hasText: regex }).first();
+
+        const address = addressContainer.locator(".address");
+        const addressActionsButton = addressContainer.locator("[id^='address-item-dropdown']");
+        const addressActionsDropdown = addressContainer.locator(".dropdown-menu");
```
frobel: The name addressActionsDropdownMenu is fitting best here I would say.

src/page-objects/storefront/AccountAddresssCreate.ts
ocavli, line 1: there is a typo in file name, it should be `AccountAddressCreate.ts`

sguder: Yeah, I wouldn't change that, as I want to deprecate the whole file anyways

ocavli (approved):

PR #592, opened 2026-02-25
fix: prevent race condition between saas setup and default sales chan…

keulinho:
…nel fixtures

Before the setup saas instance and create default sales channel fixtures might run in parallel, which might lead to race conditions with currency ids

now the sales channel fixture depends on the SaaSInstanceFixture, based on my understanding that should mean the saas instance setup fixture will run first and only after that is finished the sales channel fixture will run

### DIFF
--- a/src/fixtures/DefaultSalesChannel.ts
+++ b/src/fixtures/DefaultSalesChannel.ts
@@ -58,7 +58,9 @@ export const test = base.extend<NonNullable<unknown>, FixtureTypes>({
     ],
 
     DefaultSalesChannel: [
-        async ({ IdProvider, AdminApiContext, SalesChannelBaseConfig }, use) => {
+        // depend on `SaaSInstanceSetup` fixture so that is setup before the sales channel is created,
+        // this is required for SaaS instances to ensure the instance is ready and has the required default entities
+        async ({ IdProvider, AdminApiContext, SalesChannelBaseConfig, SaaSInstanceSetup }, use) => {
             // thread id seems to be random
 
             const { id, uuid } = IdProvider.getWorkerDerivedStableId("salesChannel");
--- a/src/page-objects/administration/RuleDetail.ts
+++ b/src/page-objects/administration/RuleDetail.ts
@@ -47,12 +47,14 @@ export class RuleDetail extends RuleCreate implements PageObject {
     public readonly conditionFilterModalCloseButtonX: Locator;
     public readonly conditionORContainer: Locator;
     public readonly adminMenuAvatar: Locator;
+    public readonly saveRuleButton: Locator;
     public readonly conditionDateRangeUseTimeOperator: Locator | undefined;
     public readonly conditionDateRangeTimezoneOperator: Locator | undefined;
 
     constructor(page: Page, instanceMeta: HelperFixtureTypes["InstanceMeta"]) {
         super(page, instanceMeta);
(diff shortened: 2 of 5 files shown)


yusufttur (approved):

ravenokko (approved):

src/fixtures/DefaultSalesChannel.ts
```diff
@@ -58,7 +58,9 @@ export const test = base.extend<NonNullable<unknown>, FixtureTypes>({
     ],
 
     DefaultSalesChannel: [
-        async ({ IdProvider, AdminApiContext, SalesChannelBaseConfig }, use) => {
+        // depend on `SaaSInstanceSetup` fixture so that is setup before the sales channel is created,
+        // this is required for SaaS instances to ensure the instance is ready and has the required default entities
+        async ({ IdProvider, AdminApiContext, SalesChannelBaseConfig, SaaSInstanceSetup }, use) => {
```
keulinho, line 63: I'll test run this a few times on a saas build so we can verify my assumptions, 

please do not merge yet

keulinho:
Saas failure atm is expected as it needs some changes in platform: https://github.com/shopware/shopware/pull/15186

src/tasks/shop-customer/Listing/CheckVisibilityInHome.ts
```diff
@@ -3,18 +3,13 @@ import type { Task } from "../../../types/Task";
 import type { FixtureTypes } from "../../../types/FixtureTypes";
 
 export const CheckVisibilityInHome = base.extend<{ CheckVisibilityInHome: Task } & FixtureTypes>({
```
keulinho, line 5: i wondered why this change didn't had any effect in platform, we have the same task there, is that intended? Can we delete one of them?

frobel: The reason to have the task here is, because we have some tests in the ats repo which need also the visibility check. So we can remove the task in platform and rely on the ATS task.

For reference see here: https://github.com/shopware/acceptance-test-suite/pull/495/changes

src/page-objects/administration/RuleDetail.ts
```diff
@@ -47,12 +47,14 @@ export class RuleDetail extends RuleCreate implements PageObject {
...
     public readonly conditionORContainer: Locator;
     public readonly adminMenuAvatar: Locator;
+    public readonly saveRuleButton: Locator;
     public readonly conditionDateRangeUseTimeOperator: Locator | undefined;
     public readonly conditionDateRangeTimezoneOperator: Locator | undefined;
 
     constructor(page: Page, instanceMeta: HelperFixtureTypes["InstanceMeta"]) {
         super(page, instanceMeta);
         this.contentView = page.locator(".sw-desktop__content");
+        this.saveRuleButton = page.getByRole("button", { name: translate("administration:rule:buttons.save") });
```
frobel: Don't need them here because of the inheritance of the `RuleCreate` page object. There is already a save button:

`this.saveButton = page.getByRole("button", { name: translate("administration:rule:buttons.save") });`

frobel (changes requested):

src/page-objects/administration/RuleDetail.ts
keulinho: good catch, i already removed that again 👍

PR #593, opened 2026-02-25
fix: update supported locales

hienha-sw:


playwright.config.ts
```diff
@@ -1,62 +1,77 @@
...
+  process.env.LANG || process.env.LANGUAGE || process.env.lang || "en-GB";
+// Normalize locale: extract locale code and map to browser locale
+const locale = envLang.split(".")[0].replace(/_/g, "-");
+const supportedLocales: Record<string, string> = {
+  "en-US": "en-US",
+  "en-GB": "en-GB",
+  "en-DE": "en-DE",
+  "de-DE": "de-DE",
+};
+const browserLocale = supportedLocales[locale] || "en-DE";
```
keulinho, line 44: this isn't correct the locale part in case of `en-DE` should be `en-GB` right? In the ATS we use the `DE` code at the end to only determine currency, 

that means this change likely is not needed at all? 🤔

PR #613, opened 2026-03-27
chore: add shopware ATS skill

vanpham-sw:


vanpham-sw:
@codex suggestions:

The draft is a bit long for a skill prompt, and it repeats the same guidance twice across SKILL.md and repo-map.md. I would trim the skill file so it stays directive and keep the repo map as the detailed reference. That will make the agent faster to read and less likely to “overlearn” one layout.

A few items look worth double-checking against the live ATS repo before you lock them in: the exact merged export files, the cleanup flag name, and the environment defaults / boot behavior. The draft is very specific there, which is good when accurate, but brittle if any path or variable differs.

I would also add a small “when to keep it simple” section. Right now the skill strongly nudges the model toward new tasks/page objects/fixtures, but it does not say enough about when a direct spec change is the right answer. That guardrail would reduce abstraction creep.

One more detail: the default_prompt using $shopware-ats may be perfect if that syntax is intentional, but I would confirm the skill runtime interprets it the way you expect. If not, I would make it plain text.

vanpham-sw:
@codex implement suggestions:

The draft is a bit long for a skill prompt, and it repeats the same guidance twice across SKILL.md and repo-map.md. I would trim the skill file so it stays directive and keep the repo map as the detailed reference. That will make the agent faster to read and less likely to “overlearn” one layout.

A few items look worth double-checking against the live ATS repo before you lock them in: the exact merged export files, the cleanup flag name, and the environment defaults / boot behavior. The draft is very specific there, which is good when accurate, but brittle if any path or variable differs.

I would also add a small “when to keep it simple” section. Right now the skill strongly nudges the model toward new tasks/page objects/fixtures, but it does not say enough about when a direct spec change is the right answer. That guardrail would reduce abstraction creep.

One more detail: the default_prompt using $shopware-ats may be perfect if that syntax is intentional, but I would confirm the skill runtime interprets it the way you expect. If not, I would make it plain text.

.codex/skills/shopware-ats/references/repo-map.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Locales
+
+- `src/locales/en/*`
+- `src/locales/de/*`
+- Keep UI assertions language-aware
+
+## Common Change Paths
+
+### Add A New Reusable Admin Flow
```
frobel, line 87: Why the headline differs to the headline of Storefront task creation?

.codex/skills/shopware-ats/references/repo-map.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Debug Failing Tests
+
+1. Re-run the narrowest spec
+2. Check environment defaults and app boot assumptions in `playwright.config.ts`
+3. Trace failure ownership through spec, task, page object, fixture, and service
+4. Fix the owning abstraction
+5. Re-run the targeted spec before broader verification
+
+### Refactor Fixtures, Tasks, Or Page Objects
```
frobel: What about Services?

.codex/skills/shopware-ats/SKILL.md
```diff
@@ -0,0 +1,57 @@
+---
+name: shopware-ats
+description: "Work effectively in the Shopware Acceptance Test Suite repository: inspect or modify Playwright specs, merged fixtures, page objects, tasks, data fixtures, service helpers, and locale-backed assertions. Use when Codex needs to debug a failing ATS test, add or refactor Shopware end-to-end coverage, trace how exported fixtures are wired together, or choose the right file to update in `tests/`, `src/fixtures/`, `src/page-objects/`, `src/tasks/`, `src/data-fixtures/`, or `src/services/`."
```
frobel: we wanna deprecate `data-fixtures` in the next major, because we have the TestDataService. So we have to consider that already in the skills

.codex/skills/shopware-ats/SKILL.md
```diff
@@ -0,0 +1,57 @@
...
+- Then open `references/repo-map.md` for merge points and edit targets.
+
+## Choose The Right Layer
+
+Pick the narrowest owner of the behavior:
+
+- `tests/`: spec flow, assertions, and coverage composition.
+- `src/page-objects/`: selectors, URLs, and page-level affordances.
+- `src/tasks/`: reusable multi-step business actions.
+- `src/data-fixtures/` or `src/services/TestDataService.ts`: generated test data and cleanup.
```
frobel: we have to avoid the usage of data-fixtures. Everything should happen with TestDataService.

.codex/skills/shopware-ats/SKILL.md
```diff
@@ -0,0 +1,57 @@
...
+Use this skill to make precise ATS changes without over-abstracting. Keep this file directive; use `references/repo-map.md` for detailed layout and examples.
+
+## Start Here
+
+- Open `package.json`, `playwright.config.ts`, and `src/index.ts`.
+- Then open `references/repo-map.md` for merge points and edit targets.
+
+## Choose The Right Layer
+
+Pick the narrowest owner of the behavior:
```
frobel, line 17: What about types?

.codex/skills/shopware-ats/references/repo-map.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Fix Setup Or Cleanup Issues
+
+1. Check `src/fixtures/TestData.ts` for cleanup gating (`ATS_SKIP_CLEANUP`)
+2. Check `src/services/TestDataService.ts` for entity creation or cleanup logic
+3. Check supporting API helpers in `src/services/`
+4. Inspect `src/data-fixtures/` only when maintaining existing legacy compatibility
+
+### Fix Broken Selectors Or Labels
+
```
ravenokko, line 126: Note sure if relevant here, exactly, but perhaps something to favor role based selectors (getByRole()). I don't want it littering our specs with .css classes.

.codex/skills/shopware-ats/SKILL.md
```diff
@@ -0,0 +1,60 @@
...
+3. Reuse is unlikely across other scenarios.
+
+Only introduce or extend page objects, tasks, fixtures, services, or shared types when the same behavior is repeated, unstable, or clearly part of shared domain language.
+
+## Implementation Workflow
+
+1. Reproduce with the smallest affected spec/test block.
+2. Search for existing nouns, flows, and `TestDataService` helpers with `rg` before creating new helpers.
+3. Trace from spec -> actor/task -> page object -> fixture/service/type and fix the owning layer.
+4. Keep established patterns (`mergeTests`, `test.extend`, actor model, locale-aware assertions).
```
ravenokko: It might struggle with 'and use `presses()`/`fillsIn()`' but only for Storefront based tests.

.codex/skills/shopware-ats/SKILL.md
```diff
@@ -0,0 +1,60 @@
...
+2. The interaction is short and readable inline.
+3. Reuse is unlikely across other scenarios.
+
+Only introduce or extend page objects, tasks, fixtures, services, or shared types when the same behavior is repeated, unstable, or clearly part of shared domain language.
+
+## Implementation Workflow
+
+1. Reproduce with the smallest affected spec/test block.
+2. Search for existing nouns, flows, and `TestDataService` helpers with `rg` before creating new helpers.
+3. Trace from spec -> actor/task -> page object -> fixture/service/type and fix the owning layer.
```
ravenokko, line 80: How can we ensure it won't mix/match admin and storefront actions? We don't want tests that navigate back and forth between the admin and storefront.

ravenokko: This will be covered in the test steps we as humans define

.codex/skills/shopware-ats/SKILL.md
ravenokko: again, we as humans will have to ensure this is applied

frobel (approved):

hienha-sw (approved):

PR #615, opened 2026-04-02
fix: cart line item stock rule selector

DennisGarding:
Because of deprecation of `ProductEntity.availableStock` and removing the rule `cartLineItemStock` we need to adjust the selector to `cartLineItemActualStock`

see PR: https://github.com/shopware/shopware/pull/16015

### DIFF
--- a/src/page-objects/administration/RuleDetail.ts
+++ b/src/page-objects/administration/RuleDetail.ts
@@ -143,7 +143,7 @@ export class RuleDetail extends RuleCreate implements PageObject {
             this.assignmentModalAddButton = this.assignmentModal.locator(".sw-button--primary").getByText(translate("administration:rule:buttons.add"));
         } else {
             this.assignmentModalAddButton = this.assignmentModal.locator(".mt-button--primary").getByText(translate("administration:rule:buttons.add"));
-        }git
+        }
         this.conditionORContainer = page.locator(".sw-condition-or-container");
     }
 
--- a/src/page-objects/administration/RuleDetail.ts
+++ b/src/page-objects/administration/RuleDetail.ts
@@ -137,8 +137,13 @@ export class RuleDetail extends RuleCreate implements PageObject {
         }
         this.conditionFilterModal = page.locator(".sw-modal__header").getByText(translate("administration:rule:text.filter"));
         this.conditionFilterModalCloseButtonX = page.locator(".sw-modal__header").getByRole("button");
-        this.conditionCartLineItemInStockOperator = page.locator(".sw-condition__condition-type-cartLineItemActualStock").locator(".sw-single-select__selection-text");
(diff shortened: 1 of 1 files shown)


MalteJanz (approved):

larskemper (approved):

jozsefdamokos (approved):

src/page-objects/administration/RuleDetail.ts
```diff
@@ -137,8 +137,8 @@ export class RuleDetail extends RuleCreate implements PageObject {
         }
         this.conditionFilterModal = page.locator(".sw-modal__header").getByText(translate("administration:rule:text.filter"));
         this.conditionFilterModalCloseButtonX = page.locator(".sw-modal__header").getByRole("button");
-        this.conditionCartLineItemInStockOperator = page.locator(".sw-condition__condition-type-cartLineItemStock").locator(".sw-single-select__selection-text");
-        this.conditionCartLineItemInStockValue = page.locator(".sw-condition__condition-type-cartLineItemStock").getByRole("textbox");
+        this.conditionCartLineItemInStockOperator = page.locator(".sw-condition__condition-type-cartLineItemActualStock").locator(".sw-single-select__selection-text");
+        this.conditionCartLineItemInStockValue = page.locator(".sw-condition__condition-type-cartLineItemActualStock").getByRole("textbox");
```
frobel: We have to make sure that the locator is still compatible with lower versions like 6.7.3.x etc.

So we have one possibility:

1. Introduce a version constraint like:

```
if (satisfies(instanceMeta.version, "<6.7.9.0")) {
[old selector value]
} else {
[new selector value]
}
```

frobel (changes requested):

src/page-objects/administration/RuleDetail.ts
MalteJanz: Btw: is this documented somewhere? Also whats the min version the ATS supports? 🤔 

Also do you run the latest ATS against older shopware versions (the min version that you support) in CI to test against issues like this automatically?

I'm just curious because I wondered about this myself in the past 🙈

DennisGarding: Done

frobel: @MalteJanz the min version is the latest LTS version of 6.6.x, then we are testing against the first major 6.7.0.0 and last major. 

We run some page-objects locators tests in the current repo and the full tests in the platform repo (6.6.x, 6.7.).

To avoid those kinds of issues you mentioned, we have to provide tests in the current repo. Unfortunately, we save time with that and use the shopware/platform repo for testing our locators. That is the reason why we are thinking of moving that project to our mono repo, that we save time for writing tests in both repositories.

frobel (approved):

PR #617, opened 2026-04-08
fix: tax price locator account order

mstegmeyer:
Fetching the tax price in the account order overview is failing due to a fixed bug (shopware/shopware#15689).
https://github.com/shopware/saas/actions/runs/23934265693

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
         "format": "oxfmt --write .",
         "typecheck": "tsgo --noEmit",
         "build": "unbuild",
-        "prepare": "lefthook install",
+        "prepare": "git rev-parse --is-inside-work-tree >/dev/null 2>&1 && lefthook install || true && npm run build",
         "prepublishOnly": "npm run build"
     },
     "dependencies": {
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
         "format": "oxfmt --write .",
         "typecheck": "tsgo --noEmit",
         "build": "unbuild",
-        "prepare": "git rev-parse --is-inside-work-tree >/dev/null 2>&1 && lefthook install || true && npm run build",
+        "prepare": "lefthook install",
         "prepublishOnly": "npm run build"
     },
     "dependencies": {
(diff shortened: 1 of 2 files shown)


src/page-objects/storefront/AccountOrder.ts
```diff
@@ -46,9 +46,7 @@ export class AccountOrder extends BaseAccount {
             name: new RegExp(`${translate("storefront:account:orders.expand")}|${translate("storefront:account:orders.showDetails")}`),
         });
         const orderImage = orderItem.locator(".line-item-img-link");
-        const taxPrice = orderItem.locator(
-            `dt:text-matches('${translate("storefront:account:orders.includeVat")} [0-9]\\+\\?${translate("storefront:account:orders.vatSuffix")}') + dd`
-        );
+        const taxPrice = orderItem.locator(`dt:text-matches(${JSON.stringify(this.buildTaxPricePattern())}) + dd`);
```
frobel, line 49: What about merchants using the Page Object in 6.6.x ...would the test fail then because the locator would result in an unknown pattern?

frobel (changes requested):

src/page-objects/storefront/AccountOrder.ts
mstegmeyer: I don't understand. The snippet is coming from inside the ATS, not from Shopware, where it can't exist. The new pattern validates for "is either one available", so `incl\.|plus`. Shouldn't that suffice for all versions?

package.json
```diff
@@ -35,7 +35,7 @@
         "format": "oxfmt --write .",
         "typecheck": "tsgo --noEmit",
         "build": "unbuild",
-        "prepare": "lefthook install",
+        "prepare": "git rev-parse --is-inside-work-tree >/dev/null 2>&1 && lefthook install || true && npm run build",
```
mstegmeyer: will be reverted

En0Ma1259 (approved):

Berkantdoku (approved):

cyl3x (approved):

src/page-objects/storefront/AccountOrder.ts
ravenokko: Tested it with `trunk` and `6.6.x` for both `CheckoutWithDifferentTaxRates` and `TaxFreeCartForCurrency`.  Works like a charm 👍

ravenokko (approved):

PR #624, opened 2026-04-16
chore(deps): replace image-js with fast-png

soulcodex:
## Summary
- replace the internal image encoding path with `fast-png`
- remove `image-js` and its `skia-canvas` / `follow-redirects` dependency chain
- add a small contract test for the image helper PNG output

## Why
This library only used `image-js` to create synthetic RGB images and encode them as PNGs for fixture/media uploads. Carrying a broader image-processing stack plus the optional native `skia-canvas` chain was unnecessary for that narrow use case.

This PR also addresses the Dependabot alert on this repository for the vulnerable transitive chain involving `follow-redirects`.

## Safety approach
To reduce migration risk, I used a temporary local golden-master check before and after the swap. It was not committed to the repo.

The golden-master comparison checked the behaviors we actually care about:
- valid PNG signature
- matching width and height
- 3 RGB channels at 8-bit depth
- identical decoded pixel values for deterministic solid-color images
- successful decode of random-image output

## Benefits
- removes an unnecessary dependency chain from the published library
- eliminates `image-js`, `skia-canvas`, and `follow-redirects` from the installed tree
- addresses the Dependabot alert at the root cause level instead of pinning a transitive override
- keeps the public helper API unchanged
- leaves behind a narrow committed contract test instead of brittle binary snapshots

## Validation
- `npm ls image-js`
- `npm ls skia-canvas`
- `npm ls follow-redirects`
- `npm run lint`
- `npm run typecheck`
- `npm run build`
- `npx playwright test tests/ImageHelper.spec.ts --project chromium`


ravenokko:
Pipeline is green but to be clear - will this also work with `6.6.x`?  Our ATS package needs to be backwards compatible.

soulcodex:
> Pipeline is green but to be clear - will this also work with `6.6.x`? Our ATS package needs to be backwards compatible.

Hi @ravenokko,

To be honest I'm new at the company so I don't know what refers you with "ATS package". Could you point me out to the intended repo ❓Do you think this is something I can test on my own or I would need help from someone at testing team ❓  

Thanks in advance

ravenokko (approved):
Tested with 6.6.x and it's working 👍 Update the package.json since we released a new version and you're all set!

jc-sw-git (approved):

PR #648, opened 2026-06-19
feat: add the create newsletter recipient request

suri-ta777:
**Summary** 
- adding the support func for create newsletter recipient in the TestDataService 
- adding the product line items locator for checkout confirm and off canvas cart page

**Test Scenario Detail**
1. Admin creates promotion with rule Newsletter recipient = Yes (will automatically receive discount - 10% at checkout)
2. Shop user:
- registers the newsletter recipient 
- adds product to cart 
- validate promotion applied on checkout + confirm page and creating order

Related to test `CheckoutWithPromotionRule.spec.ts` in PR [17408](https://github.com/shopware/shopware/pull/17408)

### DIFF
--- a/src/page-objects/storefront/CheckoutConfirm.ts
+++ b/src/page-objects/storefront/CheckoutConfirm.ts
@@ -36,10 +36,11 @@ export class CheckoutConfirm implements PageObject {
     /**
      * Product details
      */
+    public readonly cartLineItemImages: Locator;
+    public readonly page: Page;
     public readonly confirmProductTable: Locator;
     public readonly productLineItems: Locator;
     public readonly promotionLineItems: Locator;
-    public readonly page: Page;
 
     constructor(page: Page) {
         this.page = page;
--- a/src/page-objects/storefront/CheckoutConfirm.ts
+++ b/src/page-objects/storefront/CheckoutConfirm.ts
@@ -39,8 +39,6 @@ export class CheckoutConfirm implements PageObject {
     public readonly cartLineItemImages: Locator;
     public readonly page: Page;
     public readonly confirmProductTable: Locator;
-    public readonly productLineItems: Locator;
-    public readonly promotionLineItems: Locator;
 
     constructor(page: Page) {
         this.page = page;
@@ -69,12 +67,10 @@ export class CheckoutConfirm implements PageObject {
         this.legalGuaranteeNoticeLink = page.locator("#legalGuaranteeNoticeModal a");
         this.lineItemGaranLabel = page.locator(".line-item-garan-label");
         this.confirmProductTable = page.locator(".confirm-product");
-        this.productLineItems = this.confirmProductTable.locator(".line-item-product");
-        this.promotionLineItems = this.confirmProductTable.locator(".line-item-promotion");
     }
 
-    getLineItemByProductName(lineItem: Locator, productName: string): Record<string, Locator> {
-        const productLineItem = lineItem.filter({ hasText: productName });
+    getLineItemByProductName(productName: string): Record<string, Locator> {
+        const productLineItem = this.confirmProductTable.locator(".line-item-product", { hasText: productName });
         const productNameLabel = productLineItem.locator(".line-item-label");
         const productTotalPrice = productLineItem.locator(".line-item-total-price-value");
 
@@ -82,6 +78,18 @@ export class CheckoutConfirm implements PageObject {
             productLineItem: productLineItem,
             productNameLabel: productNameLabel,
             productTotalPrice: productTotalPrice,
+        };
+    }
+
+    getLineItemByPromotionName(promotionName: string): Record<string, Locator> {
+        const promotionLineItem = this.confirmProductTable.locator(".line-item-promotion", { hasText: promotionName });
+        const promotionNameLabel = promotionLineItem.locator(".line-item-label");
+        const promotionTotalPrice = promotionLineItem.locator(".line-item-total-price-value");
+
+        return {
+            promotionLineItem: promotionLineItem,
+            promotionNameLabel: promotionNameLabel,
+            promotionTotalPrice: promotionTotalPrice,
         };
     }
 
--- a/src/page-objects/storefront/CheckoutConfirm.ts
+++ b/src/page-objects/storefront/CheckoutConfirm.ts
@@ -36,7 +36,9 @@ export class CheckoutConfirm implements PageObject {
     /**
      * Product details
      */
-    public readonly cartLineItemImages: Locator;
+    public readonly confirmProductTable: Locator;
+    public readonly productLineItems: Locator;
+    public readonly promotionLineItems: Locator;
     public readonly page: Page;
 
     constructor(page: Page) {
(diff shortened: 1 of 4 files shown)


src/page-objects/storefront/CheckoutConfirm.ts
```diff
@@ -31,7 +31,9 @@ export class CheckoutConfirm implements PageObject {
     /**
      * Product details
      */
-    public readonly cartLineItemImages: Locator;
```
sguder, line 39: Why remove it? You can't just remove locators from page objects, in case anyone uses it

src/page-objects/storefront/CheckoutConfirm.ts
```diff
@@ -53,7 +55,21 @@ export class CheckoutConfirm implements PageObject {
         this.shippingStandard = page.getByLabel(translate("storefront:checkout:common.standard"));
         this.shippingExpress = page.getByLabel(translate("storefront:checkout:common.express"));
 
-        this.cartLineItemImages = page.locator(".line-item-img-link");
```
sguder, line 61: See above

sguder (changes requested):

src/services/TestDataService.ts
```diff
@@ -164,6 +182,42 @@ export class TestDataService {
         }
     }
 
+    /**
+     * Creates a newsletter recipient for a customer in the default sales channel.
+     * If a matching recipient already exists, it will be reused and registered for cleanup.
+     *
+     * @param customer - The customer whose email and profile data will be used for the newsletter recipient.
+     */
+    async createNewsletterRecipient(customer: Customer): Promise<NewsletterRecipient> {
```
sguder: For the test of the actual bug we don't have a customer yet, so we need an additional function to add a newsletter recipient with only the email address

sguder (changes requested):

src/services/TestDataService.ts
suri-ta777: This function can also support the guest customer by passing a new id with email from the test file. I updated that part from the test file already. Please help review my new code there. Thanks!

src/page-objects/storefront/CheckoutConfirm.ts
suri-ta777: my mistake working from my local. I added it back. Thanks for catching this!

src/page-objects/storefront/CheckoutConfirm.ts
suri-ta777: added this line back

src/page-objects/storefront/CheckoutConfirm.ts
```diff
@@ -54,6 +57,21 @@ export class CheckoutConfirm implements PageObject {
         this.shippingExpress = page.getByLabel(translate("storefront:checkout:common.express"));
 
         this.cartLineItemImages = page.locator(".line-item-img-link");
+        this.confirmProductTable = page.locator(".confirm-product");
+        this.productLineItems = this.confirmProductTable.locator(".line-item-product");
+        this.promotionLineItems = this.confirmProductTable.locator(".line-item-promotion");
+    }
+
+    getLineItemByProductName(lineItem: Locator, productName: string): Record<string, Locator> {
```
frobel: Why you have to provide a lineItem locator here? Could you not just take a line items and filter down to the specific one with `.filter()` like [here](https://github.com/shopware/acceptance-test-suite/blob/75e6a3c3046e5e8e3855b8a2086b30e03b68f934/src/page-objects/administration/PaymentDetail.ts#L24)

src/page-objects/storefront/OffCanvasCart.ts
```diff
@@ -66,4 +66,15 @@ export class OffCanvasCart implements PageObject {
             wishlistNotAddedButton: wishlistNotAddedButton,
         };
     }
+
+    async getLineItemByPromotionName(name: string): Promise<Record<string, Locator>> {
+        let promotionItem = this.page.locator(".line-item-promotion", { hasText: name });
```
frobel: why you use `let` as keyword here and not `const`?

src/services/TestDataService.ts
frobel: I miss the overrides here so that we can modify the date etc. (see other create functions)

frobel (changes requested):

src/page-objects/storefront/CheckoutConfirm.ts
suri-ta777: As each lineItem can be different types and will have many same assertion elements inside so i tried to make the code reusable but i updated the code based on your suggestion to align with existing code and easy to read

src/page-objects/storefront/OffCanvasCart.ts
suri-ta777: my mistake, i updated it

src/services/TestDataService.ts
suri-ta777: the overrides parameter is now added to createNewsletterRecipient func similar to other func

sguder (approved):

ravenokko (approved):

src/page-objects/storefront/OffCanvasCart.ts
```diff
@@ -68,4 +68,15 @@ export class OffCanvasCart implements PageObject {
             lineItemGaranLabel: lineItemGaranLabel,
         };
     }
+
+    async getLineItemByPromotionName(name: string): Promise<Record<string, Locator>> {
+        const promotionItem = this.page.locator(".line-item-promotion", { hasText: name });
+        const promotionLabel = promotionItem.locator(".line-item-label");
+        const promotionPrice = promotionItem.locator(".line-item-total-price-value");
+
+        return {
```
frobel, line 77: would be consistent to have `promotionItem` also as return value

src/services/TestDataService.ts
```diff
@@ -59,6 +60,23 @@ export interface SimpleLineItem {
...
+    discountScope?: string;
+    discountType?: string;
+    salesChannelId?: string;
+}
+
+export interface BasicRuleCondition {
+    type: string;
+    value: Record<string, unknown>;
+    children?: BasicRuleCondition[];
+}
```
frobel, line 78: We have  ShopwareTypes.ts for type definition. see [here](https://github.com/shopware/acceptance-test-suite/blob/trunk/src/types/ShopwareTypes.ts)

src/services/TestDataService.ts
```diff
@@ -164,6 +182,44 @@ export class TestDataService {
...
+        const recipientPayload = {
+            email: customer.email,
+            salesChannelId: this.defaultSalesChannel.id,
+            firstName: customer.firstName ?? "Test",
+            lastName: customer.lastName ?? "User",
+            hash: customer.id || hash,
+            status: "direct",
+            languageId: customer.languageId,
+            salutationId: customer.salutationId,
+            confirmedAt: new Date().toISOString(),
```
frobel, line 203:
```// Stable at any hour on any runner:
const today = new Date().toLocaleDateString('en-GB', { timeZone: 'UTC' });
```

or something similar.

src/services/TestDataService.ts
```diff
@@ -164,6 +182,44 @@ export class TestDataService {
...
+     * @param customer - The customer whose email and profile data will be used for the newsletter recipient.
+     * @param overrides - Specific data overrides that will be applied to the newsletter recipient payload.
+     */
+    async createNewsletterRecipient(customer: Customer, overrides: Partial<NewsletterRecipient> = {}): Promise<NewsletterRecipient> {
+        const hash = this.IdProvider.getIdPair();
+        const recipientPayload = {
+            email: customer.email,
+            salesChannelId: this.defaultSalesChannel.id,
+            firstName: customer.firstName ?? "Test",
+            lastName: customer.lastName ?? "User",
```
frobel, line 198: would add a hash here to Test<hash> and User<hash>

src/services/TestDataService.ts
```diff
@@ -164,6 +182,44 @@ export class TestDataService {
...
+            salutationId: customer.salutationId,
+            confirmedAt: new Date().toISOString(),
+            ...overrides,
+        };
+
+        const existingNewsletterRecipient = await this.getNewsletterRecipient({ ...customer, email: recipientPayload.email });
+        if (existingNewsletterRecipient) {
+            this.addCreatedRecord("newsletter_recipient", existingNewsletterRecipient.id);
+            return existingNewsletterRecipient;
+        }
```
frobel, line 211: You are sure you wanna "modify" or "reuse" an already existing Newsletter Recipient? I would suggest to use everytime a unique one for testing.

frobel (changes requested):

src/page-objects/storefront/OffCanvasCart.ts
suri-ta777: it's now returned the `promotionItem` as well

src/services/TestDataService.ts
suri-ta777: i added hash for them

frobel (approved):

PR #655, opened 2026-07-02
fix: set type: 'digital' in createDigitalProduct

### LINKED ISSUE #654: Fix createDigitalProduct in acceptance-test-suite to set type: 'digital'
TestDataService.createDigitalProduct() creates a product without explicitly setting type: 'digital'. This is the same invalid test data issue that caused the digital product download incident to go undetected — the test was passing while production was broken.

**File:** src/services/TestDataService.ts:246

**Fix:**

**Before**
`const product = await this.createBasicProduct(overrides, taxId, currencyId);`

**After**
`const product = await this.createBasicProduct({ type: 'digital', ...overrides }, taxId, currencyId);
`Impact: Any test using createDigitalProduct is currently testing an invalid product state that does not represent real production data.

ngocblue:
Closes #654

## Problem

`TestDataService.createDigitalProduct()` created products without explicitly setting `type: 'digital'`. Tests relying on this helper were exercising an invalid product state that does not represent real production data — the same class of invalid test data that let the digital product download incident go undetected while production was broken.

## Changes

- **`src/services/TestDataService.ts`** — `createDigitalProduct()` now passes `{ type: 'digital', ...overrides }` to `createBasicProduct()`, so digital products are created with the correct type. Placing `...overrides` last keeps caller-supplied overrides authoritative.
- **`src/types/ShopwareTypes.ts`** — added an optional `type?: string` field to the `Product` type so the new property is type-safe.

## Impact

Any test using `createDigitalProduct` now works with a valid, production-like digital product instead of a mistyped one.

## Verification

- ✅ `npm run build` completes successfully.
- ✅ Ran the tests covering `createDigitalProduct` against the platform — all pass.

### DIFF
--- a/src/services/TestDataService.ts
+++ b/src/services/TestDataService.ts
@@ -243,7 +243,7 @@ export class TestDataService {
      * @param currencyId - The uuid of the currency to use for the product pricing.
      */
     async createDigitalProduct(content = "Lorem ipsum dolor", overrides: Partial<Product> = {}, taxId = this.defaultTaxId, currencyId = this.defaultCurrencyId): Promise<Product> {
-        const product = await this.createBasicProduct(overrides, taxId, currencyId);
+        const product = await this.createBasicProduct({ type: 'digital', ...overrides }, taxId, currencyId);
         const media = await this.createMediaTXT(content);
 
         await this.assignProductDownload(product.id, media.id);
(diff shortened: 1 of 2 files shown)


ocavli (approved):

ocavli:
It is ok from my side, would be nice I you could run shopware and commercial tests if they will fail with these changes...If so needs to update them as well.

ngocblue:
> It is ok from my side, would be nice I you could run shopware and commercial tests if they will fail with these changes...If so needs to update them as well.

Yes, I searched on both platform and commercial and saw that there is only one test case call to .createDigitalProduct() on platform. I ran it with the new change from ATS, and it passed; there is no need to update.

hienha-sw (approved):

PR #662, opened 2026-07-24
fix: default digital products to maxPurchase 1 and assert quantity selector visibility

### LINKED ISSUE #654: Fix createDigitalProduct in acceptance-test-suite to set type: 'digital'

ngocblue:
## Description

`createDigitalProduct()` now defaults `maxPurchase: 1`, matching what the Administration does when a product is created as a digital type. Previously fixture-created digital products had no purchase limit, so the storefront rendered a quantity selector and allowed buying more than one — unlike real digital products.

`AddProductToCart` now derives whether the storefront should show the quantity selector (`type === "digital" && maxPurchase === 1` → hidden) and asserts it: **visible** for normal products (then fills the quantity), **hidden** for fixed-quantity digital products (adds a single unit). This mirrors the storefront's `showQuantitySelect` rule in `buy-widget-form.html.twig`.

Callers can still override via `createDigitalProduct(content, { maxPurchase: N })`.

## Verification

- `CheckoutWithDigitalProduct.spec.ts` passes (digital → selector hidden)
- `RegisteredUserBuysProduct.spec.ts` passes (normal → selector visible)

### DIFF
--- a/src/services/TestDataService.ts
+++ b/src/services/TestDataService.ts
@@ -237,13 +237,17 @@ export class TestDataService {
      * Creates a digital product with a text file as its download.
      * The product will be added to the default sales channel category if configured.
      *
+     * Digital products default to `maxPurchase: 1`, mirroring the Administration
+     * which forces this limit when a product is created as a digital type.
+     * Pass `maxPurchase` in the overrides to change it.
+     *
      * @param content - The content of the text file for the product download.
      * @param overrides - Specific data overrides that will be applied to the product data struct.
      * @param taxId - The uuid of the tax rule to use for the product pricing.
      * @param currencyId - The uuid of the currency to use for the product pricing.
      */
     async createDigitalProduct(content = "Lorem ipsum dolor", overrides: Partial<Product> = {}, taxId = this.defaultTaxId, currencyId = this.defaultCurrencyId): Promise<Product> {
-        const product = await this.createBasicProduct({ type: 'digital', ...overrides }, taxId, currencyId);
+        const product = await this.createBasicProduct({ type: "digital", maxPurchase: 1, ...overrides }, taxId, currencyId);
         const media = await this.createMediaTXT(content);
 
         await this.assignProductDownload(product.id, media.id);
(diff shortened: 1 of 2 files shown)


ravenokko (approved):

sguder (approved):

src/services/TestDataService.ts
```diff
@@ -237,13 +237,17 @@ export class TestDataService {
...
+     * Pass `maxPurchase` in the overrides to change it.
+     *
      * @param content - The content of the text file for the product download.
      * @param overrides - Specific data overrides that will be applied to the product data struct.
      * @param taxId - The uuid of the tax rule to use for the product pricing.
      * @param currencyId - The uuid of the currency to use for the product pricing.
      */
     async createDigitalProduct(content = "Lorem ipsum dolor", overrides: Partial<Product> = {}, taxId = this.defaultTaxId, currencyId = this.defaultCurrencyId): Promise<Product> {
-        const product = await this.createBasicProduct({ type: 'digital', ...overrides }, taxId, currencyId);
+        const product = await this.createBasicProduct({ type: "digital", maxPurchase: 1, ...overrides }, taxId, currencyId);
```
yusufttur, line 250: Why do we have to give max purchase 1, if it is a digital product, that should come as default.

yusufttur (changes requested):

src/services/TestDataService.ts
ngocblue: When you create a digital product in the Administration, it sends both type: "digital" and maxPurchase: 1 in the payload:
Screenshot 2026-07-24 at 17 28 30
You're right that maxPurchase isn't strictly what should drive this — but it is the field the storefront actually keys off, and it had a bug that the Inventory team fixed in [#16652](https://github.com/shopware/shopware/pull/16652). So I think it's safer to set maxPurchase in createDigitalProduct() as well, so the fixture mirrors exactly how a digital product is created from the Admin page.

yusufttur (approved):

PR #666, opened 2026-07-31
feat: reuse one admin session per worker

janbuecker:
## Why

Every test that touches the Administration creates a fresh admin user and performs a full UI login: new browser context, login form, an explicit wait for every plugin bundle, dashboard render, cache purge. That is 10–20 seconds per test on CI/SaaS, and the platform suite pays it ~90 times per run.

## What changed

- New worker-scoped **`AdminSession`** fixture: one admin user, one UI login per worker.
- **`AdminPage` stays test-scoped**, but now opens a fresh page inside the shared authenticated context. Tests keep an isolated DOM and JavaScript world — they just stop logging in.
- API mocks are installed on the **context before the first navigation** — previously the initial admin page load bypassed them.
- **Git installs work now:** `prepare` builds `dist/`, so consumers can test unreleased branches via `npm install github:shopware/acceptance-test-suite#<branch>`.
- New optional **`ATS_ID_SEED`** env to separate the worker-derived stable ids of concurrent suite runs against the same shop (e.g. Platform and Commercial as parallel CI jobs).
- Bugfix: `getWorkerDerivedStableId` returned the bare worker index as its short id, so derived names/emails (`customer_0@example.com`) collide globally between any two suites pointed at one shop. The id is now derived from the seed hash.
- Bugfix: the admin user struct sent `timezone`; the schema field is `timeZone`, so the timezone was silently never applied.

## What this means for consumers — please review these two semantic changes

1. **Tests within one worker share one admin user.** Server-side per-user state (`user_config`, grid preferences) persists across the tests of a worker. Tests that need a fresh, specific, or permission-restricted user should create one and log in via `createNewAdminPageContext` + `loginToAdministration` — an audit of this repo, shopware/platform, and SwagCommercial found every existing ACL/fresh-user test already does exactly that.
2. **The `context` alias now returns the shared worker context.** No current test mutates or closes it. On failure Playwright disposes the worker, so retries always get a completely fresh session.

## Measured

- Local trunk shop: admin fixture per test 1.3–2.7 s → 0.36–1.7 s.
- shopware/saas CI pipeline at identical worker count: Platform suite 16.6 → 9.5 min, Commercial 18.9 → ~14 min — reproduced across multiple runs, with every remaining failure control-tested as pre-existing.

Validated end to end via shopware/saas#748, which installs this branch through the new git-ref mechanism.

https://claude.ai/code/session_01MZJW1coXncj5CUE3AtpyLM

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
         "format": "oxfmt --write .",
         "typecheck": "tsgo --noEmit",
         "build": "unbuild",
-        "prepare": "lefthook install",
+        "prepare": "lefthook install && npm run build",
         "prepublishOnly": "npm run build"
     },
     "dependencies": {
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
         "format": "oxfmt --write .",
         "typecheck": "tsgo --noEmit",
         "build": "unbuild",
-        "prepare": "lefthook install && npm run build",
+        "prepare": "(lefthook install || true) && npm run build",
         "prepublishOnly": "npm run build"
     },
     "dependencies": {
--- a/src/fixtures/DefaultSalesChannel.ts
+++ b/src/fixtures/DefaultSalesChannel.ts
@@ -4,7 +4,7 @@ import type { Customer, SalesChannel } from "../types/ShopwareTypes";
 import type { components } from "@shopware/api-client/admin-api-types";
 import { getCountryAddressData } from "../services/ShopwareDataHelpers";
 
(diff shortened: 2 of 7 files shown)


frobel:
Glad to see that improvement :) I also had that improvement in mind to differentiate between Admin Session and Admin Login. Now we can safe 20sec of session creation and login procedure per test. Nice work!

frobel (approved):

yusufttur (approved):

PR #674, opened 2026-08-12
fix: wait for storefront JavaScript compilation

keulinho:
## What?

Require the theme readiness check to validate the main storefront JavaScript as well as CSS. Compile the assigned theme synchronously and remove the SaaS-only cache-clear polling.

## Why?

A page can expose compiled CSS before its storefront JavaScript entry is available. Treating that state as ready lets tests navigate to markup whose plugins cannot initialize.


ravenokko (approved):

keulinho:
can be closed, was not the root cause for flakiness, it was this: https://github.com/shopware/shopware/pull/19279

PR #678, opened 2026-08-17
feat: add function to set primary order references

sguder:
Adds order references to an order after creation regarding https://github.com/shopware/shopware/pull/6864


src/services/TestDataService.ts
```diff
@@ -736,6 +737,54 @@ export class TestDataService {
...
+            const transactionData = await transactionResponse.json();
+            transaction = transactionData.data?.[0];
+        }
+        if (deliveryResponse.ok()) {
+            const deliveryData = await deliveryResponse.json();
+            delivery = deliveryData.data?.[0];
+        }
+
+        if (!transaction || !delivery) {
+            throw new Error(`Order ${order.id} does not contain a transaction and delivery`);
```
frobel: It is an OR instead of an AND (see if-function definition)

frobel (changes requested):

src/services/TestDataService.ts
sguder: Or is correct. If one of them is missing, throw an error.

sguder: Think I improved it an other way. Less confusing, more informative..

frobel (approved):

frobel (approved):

ravenokko (approved):

PR #691, opened 2026-08-28
fix: assign domain currencies to sales channels

keulinho:
## What does this PR do?

Assign the currency selected for a generated sales-channel domain before creating the domain. This keeps the fixture valid when a platform version validates domain currencies against the sales channel assignment.

## How was it tested?

- `npm run typecheck`
- `npm run lint`

### DIFF
--- a/src/services/TestDataService.ts
+++ b/src/services/TestDataService.ts
@@ -1102,6 +1102,8 @@ export class TestDataService {
 
         const salesChannelDomainStruct = this.getSalesChannelDomainStruct(salesChannelId, currencyId, languageId, snippetSetId, overrides);
 
+        await this.assignSalesChannelCurrency(salesChannelId, currencyId);
+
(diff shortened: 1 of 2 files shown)


src/services/TestDataService.ts
```diff
@@ -3017,6 +3017,10 @@ export class TestDataService {
             currencyId: currencyId,
             languageId: languageId,
             snippetSetId: snippetSetId,
+            salesChannel: {
+                id: salesChannelId,
+                currencies: [{ id: currencyId }],
```
keulinho, line 3022: Before we assigned a currency to the domain that is not available in the sales channel, the fact that we didn't enforce that was clearly a bug 🙈 when we fixed the bug the ATS pipeline went red, with this change it's fixed: core PR https://github.com/shopware/shopware/pull/19836

yusufttur (approved):

frobel (approved):

PR #698, opened 2026-09-10
fix(flow): use mt-tabs locators for flow detail tabs behind v6.8.0.0

### LINKED ISSUE #18182: Stabilize acceptance test: Settings/FlowBuilderCreate.spec.ts (major)
Parent epic: #17559

Failing acceptance spec on the next major (Platform), reproducible on PHP **8.2 / 8.4 / 8.5** in Nightly [run 28987958476](https://github.com/shopware/shopware/actions/runs/28987958476) (2026-07-09, `trunk`@`ee03bb09`).

**Spec:** [`Settings/FlowBuilderCreate.spec.ts`](https://github.com/shopware/shopware/blob/trunk/tests/acceptance/tests/Settings/FlowBuilderCreate.spec.ts) · [Currents result](https://app.currents.dev/instance/34eaee6bc3525acc)

### Failing tests (1)
- `4:1` — As an admin user, I want to create a new flow · [Currents](https://app.currents.dev/i/https://app.currents.dev/i/34eaee6bc3525acc/test/5fa206d1d9f4d1b966d5-cbd49464d0ce1feb53ee)

### Task
- [ ] Reproduce the failure locally / in CI.
- [ ] Root-cause: broken spec vs. real product regression on major.
- [ ] Fix the spec, or coordinate the fix with the responsible dev team.
- [ ] Confirm the spec passes on major (PHP 8.2 / 8.4 / 8.5).

_Source: Nightly run 28987958476 � Currents run `0182569cf1d33ee7`._

### LINKED ISSUE #18183: Stabilize acceptance test: Settings/FlowBuilderTemplates.spec.ts (major)
Parent epic: #17559

Failing acceptance spec on the next major (Platform), reproducible on PHP **8.2 / 8.4 / 8.5** in Nightly [run 28987958476](https://github.com/shopware/shopware/actions/runs/28987958476) (2026-07-09, `trunk`@`ee03bb09`).

**Spec:** [`Settings/FlowBuilderTemplates.spec.ts`](https://github.com/shopware/shopware/blob/trunk/tests/acceptance/tests/Settings/FlowBuilderTemplates.spec.ts) · [Currents result](https://app.currents.dev/instance/d1c1a88fcbb3560d)

### Failing tests (1)
- `4:1` — As an admin, I want to create new flows from templates, so that I can easily create new ones based on the default flows. · [Currents](https://app.currents.dev/i/https://app.currents.dev/i/d1c1a88fcbb3560d/test/41be2147d4c086417eb0-21888ef8ae616b083acc)

### Task
- [ ] Reproduce the failure locally / in CI.
- [ ] Root-cause: broken spec vs. real product regression on major.
- [ ] Fix the spec, or coordinate the fix with the responsible dev team.
- [ ] Confirm the spec passes on major (PHP 8.2 / 8.4 / 8.5).

_Source: Nightly run 28987958476 � Currents run `0182569cf1d33ee7`._

jc-sw-git:
## Summary
- `sw-flow-detail.html.twig` switches from `sw-tabs` to `mt-tabs` once the `v6.8.0.0` feature flag is active, and `mt-tabs` no longer renders the `sw-flow-detail__tab-general`/`sw-flow-detail__tab-flow` classes `FlowBuilderCreate`/`FlowBuilderDetail` locators relied on. This broke `Settings/FlowBuilderCreate.spec.ts`, `Settings/FlowBuilderTemplates.spec.ts` and (indirectly) `Settings/FlowValidation.spec.ts` on the major nightly (shopware/shopware#18182, #18183, #18184).
- Both page objects now branch on `instanceMeta.features["V6_8_0_0"]` (the live flag state from `/api/_action/feature-flag`, so it also works under `FEATURE_ALL=major` on a still-6.7-versioned trunk) and use the `mt-tabs` item's stable `data-item-name` (Vue Router name) instead of the removed class.
- Added `FlowBuilderTemplates.searchLineItemByFlowName()` to replace a duplicated local search+wait helper that lived in the shopware/shopware spec, and it exposes the `createFlowLink`/`templateDetailLink` locators the spec was previously re-deriving via raw CSS classes.

## Test plan
- [x] `npm run typecheck`
- [x] `npm run build`
- [x] Verified against a local trunk with `FEATURE_ALL=major` (mt-tabs branch) and with `FEATURE_ALL` unset (legacy sw-tabs branch): `Settings/FlowBuilderTemplates.spec.ts` passes in both modes after syncing this build into shopware/shopware's `tests/acceptance/node_modules/@shopware-ag/acceptance-test-suite/dist/`.

Companion PR in shopware/shopware (test refactor, depends on this being released): TBD

### DIFF
--- a/src/page-objects/administration/FlowBuilderCreate.ts
+++ b/src/page-objects/administration/FlowBuilderCreate.ts
@@ -58,9 +58,16 @@ export class FlowBuilderCreate implements PageObject {
         this.saveButton = page.locator(".sw-flow-detail__save");
         this.header = page.locator("h2");
         this.smartBarHeader = page.locator(".smart-bar__header");
-        this.generalTab = page.locator(".sw-flow-detail__tab-general");
         this.triggerSelectField = page.locator(".sw-flow-detail-flow__trigger-card").getByRole("textbox");
-        this.flowTab = page.locator(".sw-tabs__content").locator(".sw-flow-detail__tab-flow");
+        //TODO: After 6.8.0.0 release if condition needs to be replaced with 'satisfies(instanceMeta.version, ">=6.8")'
+        if (instanceMeta.features["V6_8_0_0"]) {
+            this.generalTab = page.locator('.mt-tabs__item[data-item-name="sw.flow.create.general"]');
+            this.flowTab = page.locator('.mt-tabs__item[data-item-name="sw.flow.create.flow"]');
+        } else {
+            this.generalTab = page.locator(".sw-flow-detail__tab-general");
+            this.flowTab = page.locator(".sw-tabs__content").locator(".sw-flow-detail__tab-flow");
+        }
+        //TODO: After 6.8.0.0 release this condition can be removed. Only the else branch is needed then
         if (satisfies(instanceMeta.version, "<6.7")) {
             this.modalAddButton = page.locator(".sw-button--primary").getByText(translate("administration:flowBuilder:create.addAction"));
         } else {
--- a/src/page-objects/administration/FlowBuilderTemplates.ts
+++ b/src/page-objects/administration/FlowBuilderTemplates.ts
@@ -23,8 +23,6 @@ export class FlowBuilderTemplates extends FlowBuilderListing implements PageObje
         };
     }
 
-    // Waits for the search-triggered /api/search/flow-template request matching searchTerm to
-    // resolve before resolving the row, so the row lookup does not race the grid re-render.
     async searchLineItemByFlowName(searchTerm: string, flowName: string) {
         const searchResponse = this.page.waitForResponse((response) => {
             if (!response.url().includes("/api/search/flow-template") || response.request().method() !== "POST") {
--- a/src/page-objects/administration/FlowBuilderTemplates.ts
+++ b/src/page-objects/administration/FlowBuilderTemplates.ts
@@ -29,6 +29,10 @@ export class FlowBuilderTemplates extends FlowBuilderListing implements PageObje
                 return false;
             }
 
+            if (!response.ok()) {
+                return false;
+            }
+
             const requestData = response.request().postDataJSON() as { term?: string } | null;
 
             return requestData?.term === searchTerm;
--- a/src/page-objects/administration/FlowBuilderTemplates.ts
+++ b/src/page-objects/administration/FlowBuilderTemplates.ts
@@ -22,4 +22,23 @@ export class FlowBuilderTemplates extends FlowBuilderListing implements PageObje
             templateDetailLink: templateDetailLink,
         };
     }
+
+    // Waits for the search-triggered /api/search/flow-template request matching searchTerm to
+    // resolve before resolving the row, so the row lookup does not race the grid re-render.
+    async searchLineItemByFlowName(searchTerm: string, flowName: string) {
+        const searchResponse = this.page.waitForResponse((response) => {
+            if (!response.url().includes("/api/search/flow-template") || response.request().method() !== "POST") {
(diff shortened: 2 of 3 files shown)


src/page-objects/administration/FlowBuilderCreate.ts
```diff
@@ -58,9 +58,16 @@ export class FlowBuilderCreate implements PageObject {
         this.saveButton = page.locator(".sw-flow-detail__save");
         this.header = page.locator("h2");
         this.smartBarHeader = page.locator(".smart-bar__header");
-        this.generalTab = page.locator(".sw-flow-detail__tab-general");
         this.triggerSelectField = page.locator(".sw-flow-detail-flow__trigger-card").getByRole("textbox");
-        this.flowTab = page.locator(".sw-tabs__content").locator(".sw-flow-detail__tab-flow");
+        //TODO: After 6.8.0.0 release if condition needs to be replaced with 'satisfies(instanceMeta.version, ">=6.8")'
```
sguder, line 62: How do we keep track of this, so it is not falling under the table?

sguder (approved):

src/page-objects/administration/FlowBuilderCreate.ts
jc-sw-git: In this particular case the test will actually fail once the 6.8 is released. Though, in general we need to define an approach for it. As far as I am aware we don't have such a tracking flow in place yet, so I've added it as a discussion point to our next QOps weekly

src/page-objects/administration/FlowBuilderTemplates.ts
```diff
@@ -22,4 +22,21 @@ export class FlowBuilderTemplates extends FlowBuilderListing implements PageObje
...
+            const requestData = response.request().postDataJSON() as { term?: string } | null;
+
+            return requestData?.term === searchTerm;
+        });
+
+        await this.searchBar.fill(searchTerm);
+        await searchResponse;
+
+        return this.getLineItemByFlowName(flowName);
+    }
```
yusufttur, line 45: Consider adding HTTP status code validation to ensure the API call succeeded, not just that it matched the expected URL and term. This follows the established pattern in the codebase (see `UploadMedia.ts`, `BulkEditProducts.ts`, and `SaveProduct.ts`).

Add `response.ok()` check inside the predicate:

```typescript
async searchLineItemByFlowName(searchTerm: string, flowName: string) {
    const searchResponse = this.page.waitForResponse((response) => {
        if (!response.url().includes("/api/search/flow-template") || response.request().method() !== "POST") {
            return false;
        }

        if (!response.ok()) {  // ✅ Add this check
            return false;
        }

        const requestData = response.request().postDataJSON() as { term?: string } | null;
        return requestData?.term === searchTerm;
    });

    await this.searchBar.fill(searchTerm);
    await searchResponse;

    return this.getLineItemByFlowName(flowName);
}
```

**Why this matters:**

Without the response.ok() check, the method will wait for and return a matching API response regardless of whether it succeeded (2xx) or failed (4xx/5xx). This means a failed search API call would still proceed to getLineItemByFlowName(), potentially causing confusing test failures downstream instead of failing immediately with a clear API error. This is a defensive practice used consistently throughout the codebase to catch API errors early.

yusufttur (approved):
A small API status verification would be nice to have.

src/page-objects/administration/FlowBuilderTemplates.ts
jc-sw-git: Good catch, applied in 27ae636 — matches the UploadMedia.ts pattern.

PR #703, opened 2026-09-14
fix(flow): give create-flow header assertion its own timeout

yusufttur:
## Summary
The `CreateFlow` task's assertion that the smart-bar header shows "New flow" right after clicking the create button was failing intermittently on loaded/shared environments (see `Settings/FlowBuilderCreate.spec.ts`, tracked in shopware/shopware#15749).

**Root cause, confirmed against the Administration source (not just test-side guessing):** `createFlowButton` is rendered as:
```html
<mt-button class="sw-flow-list__create" variant="primary"
  :disabled="!acl.can('flow.creator') || undefined"
  @click="$router.push({ name: 'sw.flow.create' })">
```
`acl.can()` synchronously reads `Shopware.Store.get('session').userPrivileges`. On a loaded/shared environment that state can still be settling right after landing on the listing page, so the button is genuinely disabled for a moment — clicking it then silently no-ops instead of navigating. This is a real (if narrow) app-side timing window, not purely a Playwright-is-faster-than-a-human artifact; a route-transition/pointer-events hypothesis was checked and ruled out (no `<transition>` wraps the router-view, no relevant `pointer-events: none` in admin CSS).

**Fix:**
- Wait for `createFlowButton` to actually be enabled before clicking.
- Attach the `waitForURL` listener before the click (not after), so a fast route change can't be missed.

## Evidence
A fresh nightly PaaS ATS run reproduces the exact failure this PR addresses, on the currently published `@shopware-ag/acceptance-test-suite@12.10.2` (pre-fix):
https://github.com/shopware-redstone/nightly-paas-ats/actions/runs/34798740404/job/103845141399
```
Error: expect(locator).toHaveText(expected) failed
Locator:  locator('.smart-bar__header')
Expected: "New flow"
Received: "Flow Builder  (146) "
Timeout:  5000ms
```

## Test plan
- [x] `npm run lint`
- [x] `npm run typecheck`
- [ ] Re-run `tests/Settings/FlowBuilderCreate.spec.ts` in nightly-paas-ats once this package version is published/consumed there, to confirm the fix holds under the same conditions as the run linked above


sguder (approved):

jc-sw-git (changes requested):
One thing I was curious about while reading through this: do we have a sense of whether this is purely a test-side thing — Playwright clicking faster/more precisely than any human ever would, hitting a tiny timing window — or whether it could point to a small reliability gap in the Admin app itself that real users on a slower device/connection might occasionally hit too? No strong opinion either way, just think it's worth a quick gut-check before we settle on "retry the click," since that would change where we'd want the fix to live long-term.

If it looks like a test-side thing, a couple of ideas that might get us a bit closer to the actual cause (totally optional, happy to discuss):

- Maybe a CSS-transition timing window? If the button briefly has `pointer-events: none` during the route-change animation, that would explain a click silently doing nothing. As a quick local debug check, emulating `reducedMotion: "reduce"` on the page would confirm/rule this out (if the flakiness disappears with animations off, that's a strong signal). I wouldn't want to actually ship that in the test config though — it'd mean testing a browser state (`prefers-reduced-motion: reduce`) that barely exists in real installations out there, so it'd just be masking the timing window rather than testing around it. If the hypothesis holds, waiting for the actual transition to end seems like the better fix, e.g.:
  ```ts
  await AdminFlowBuilderListing.page.waitForFunction(
    (el) => getComputedStyle(el).pointerEvents !== "none",
    await AdminFlowBuilderListing.createFlowButton.elementHandle(),
  );
  await AdminFlowBuilderListing.createFlowButton.click();
  ```
- Or maybe `createFlowButton` is briefly disabled/covered right after landing on the listing (e.g. a pending ACL check)? A quick `toBeEnabled()` check before the click might make it deterministic:
  ```ts
  await ShopAdmin.expects(AdminFlowBuilderListing.createFlowButton).toBeEnabled();
  await AdminFlowBuilderListing.createFlowButton.click();
  ```
- Also just noticed the URL-wait is attached only after the click fires — mirroring the `flowActionsLoaded` pattern a few lines above (listen first, then act) might close one more small gap:
  ```ts
  const navigatedToCreate = AdminFlowBuilderCreate.page.waitForURL((url) => url.hash.includes("/sw/flow/create/"));
  await AdminFlowBuilderListing.createFlowButton.click();
  await navigatedToCreate;
  ```

I do think these are worth actually checking rather than just a "nice to have" — if one of them holds up, it would let us wait for a real, checkable condition instead of blindly retrying against a timeout, which seems like the better fix for the flakiness itself, not just a stylistic preference. If it turns out none of them explain it and the click genuinely, unpredictably no-ops even when actionable, the retry is a reasonable fallback

yusufttur:
Good gut-check, and both your leads panned out — no retry needed after all.

I checked hypothesis 1 (CSS/transition `pointer-events`) directly against the Administration source: no `<transition>` wraps the router-view, and the only `pointer-events: none` in the admin CSS is scoped to the collapsed side-nav flyout — unrelated. Ruled out.

Hypothesis 2 held, though not from an async loading flag — it's simpler than that. `createFlowButton`'s template is:
```html
<mt-button class="sw-flow-list__create" variant="primary"
  :disabled="!acl.can('flow.creator') || undefined"
  @click="$router.push({ name: 'sw.flow.create' })">
```
`acl.can()` reads `Shopware.Store.get('session').userPrivileges` synchronously, but that store state can still be settling right after landing on the listing page on a loaded environment — so the button is genuinely disabled for a moment, and the click silently no-ops. Real timing window, not just Playwright being faster than a human.

Applied both of your snippets as-is:
```ts
await ShopAdmin.expects(AdminFlowBuilderListing.createFlowButton).toBeEnabled();
const navigatedToCreate = AdminFlowBuilderCreate.page.waitForURL((url) => url.hash.includes("/sw/flow/create/"));
await AdminFlowBuilderListing.createFlowButton.click();
await navigatedToCreate;
```
Dropped the `expect(...).toPass()` retry I'd added earlier — with the actual condition now waited on deterministically, it was redundant belt-and-suspenders.

src/tasks/shop-admin/Flow/CreateFlow.ts
```diff
@@ -11,7 +11,16 @@ export const CreateFlow = base.extend<{ CreateFlow: Task }, FixtureTypes>({
...
+                // createFlowButton stays disabled until the admin's ACL privileges are loaded
+                // (see acl.can('flow.creator') on the button in the administration), which can still
+                // be settling right after landing on this page on a loaded/shared environment; clicking
+                // it while disabled silently no-ops instead of navigating (tracked in shopware/shopware#15749).
+                await ShopAdmin.expects(AdminFlowBuilderListing.createFlowButton).toBeEnabled();
+                // The listener is attached before the click so a fast route change can't be missed.
+                const navigatedToCreate = AdminFlowBuilderCreate.page.waitForURL((url) => url.hash.includes("/sw/flow/create/"));
                 await AdminFlowBuilderListing.createFlowButton.click();
+                await navigatedToCreate;
+                await ShopAdmin.expects(AdminFlowBuilderCreate.page.locator(".sw-skeleton")).toHaveCount(0);
```
jc-sw-git, line 23: I think, this should be part of the page object, wdyt?

src/tasks/shop-admin/Flow/CreateFlow.ts
```diff
@@ -11,7 +11,16 @@ export const CreateFlow = base.extend<{ CreateFlow: Task }, FixtureTypes>({
                 // Listens for the flow-actions.json response to ensure the action dropdown is populated once it is needed.
                 const flowActionsLoaded = AdminFlowBuilderCreate.page.waitForResponse((response) => response.url().includes("/_info/flow-actions.json") && response.ok());
 
+                // createFlowButton stays disabled until the admin's ACL privileges are loaded
+                // (see acl.can('flow.creator') on the button in the administration), which can still
+                // be settling right after landing on this page on a loaded/shared environment; clicking
+                // it while disabled silently no-ops instead of navigating (tracked in shopware/shopware#15749).
+                await ShopAdmin.expects(AdminFlowBuilderListing.createFlowButton).toBeEnabled();
+                // The listener is attached before the click so a fast route change can't be missed.
+                const navigatedToCreate = AdminFlowBuilderCreate.page.waitForURL((url) => url.hash.includes("/sw/flow/create/"));
```
jc-sw-git, line 20: That's just nitpicking and I really don't mind, if we leave it as it is but personally I think, it would read better, when Line 11-12 would come after line 18. That way we would check first if the button is enabled and prepare the listener afterwards exactly when we need them to listen for the first time, which is in both cases right before the click.

Also, and maybe that's just me, but I am not a huge fan of the huge comment fields, that Claude creates. I think, it is just too much info and context, that is not always necessary to have written down for eternity in the code. It's important for the context of the PR to make the reviewing a bit better understandable. Personally, I wouldn't need the comments in Line14-17 at all.

src/tasks/shop-admin/Flow/CreateFlow.ts
jc-sw-git: I think, this should be part of the page object, wdyt?

src/tasks/shop-admin/Flow/CreateFlow.ts
yusufttur: I am also not a fan of this, asked my AI friend before pushing last time and it resisted and suggested to keep, but now I won't listen him. 😄

src/tasks/shop-admin/Flow/CreateFlow.ts
yusufttur: created a locator for the skeleton element.

jc-sw-git (approved):

PR #706, opened 2026-09-16
feat(page-objects): add customer group tax display and price basis radio locators

mstegmeyer:
The customer group detail page in the Administration replaces its single gross/net toggle with two radio groups, tax display and price basis ([shopware/shopware#19474](https://github.com/shopware/shopware/pull/19474)). This adds locators for the four options so platform specs can assert them without inlining raw selectors.

The change is purely additive: `customerGroupGrossTaxDisplay` / `customerGroupNetTaxDisplay` keep pointing at the old markup. The usual alternatives do not work here, because the 6.7 maintenance line and platform trunk cannot be told apart — both report `6.7.9999999-dev`, so a `satisfies` range matches both, and the template change is not behind a feature flag either. Keeping both sets of locators lets maintenance branches and trunk each use the ones matching their markup, with no risk of silently breaking the other.

The new locators resolve to the `input[type="radio"]` elements via the field-name ids that `sw-radio-field` generates, matching the file's existing locator style.


jc-sw-git (approved):

src/page-objects/administration/CustomerGroupCreate.ts
```diff
@@ -32,6 +42,10 @@ export class CustomerGroupCreate implements PageObject {
         this.customerGroupNameField = page.locator("#sw-field--customerGroup-name");
         this.customerGroupGrossTaxDisplay = page.locator("#sw-field--castedValue-0");
         this.customerGroupNetTaxDisplay = page.locator("#sw-field--castedValue-1");
+        this.customerGroupTaxDisplayGross = page.locator("#sw-field--customerGroup-displayGross-0");
+        this.customerGroupTaxDisplayNet = page.locator("#sw-field--customerGroup-displayGross-1");
```
frobel, line 46: Why is it called `displayGross-1` instead of  for example `displayNet-0`...is that by design?

PR #708, opened 2026-09-22
fix(bulk-edit): update customer custom-field selectors for mt-tabs

### LINKED ISSUE #19236: Bulk edit custom fields: "Change" toggle missing on 6.8 (mt-tabs)
### Shopware Version
6.8.0.0 (trunk, `FEATURE_ALL=major`)

### Affected area / extension
Platform(Default)

### Actual behaviour
In bulk edit (Customers and Products), the Custom fields section no longer shows the per-field "Change: <field>" toggle on 6.8. Only the raw field renders, so there is no way to select which custom field to apply to the selected records. Every other bulk-edit section (Tags, Customer group, …) still shows its "Change" toggle.

Image

Short root cause: `sw-bulk-edit-custom-fields` adds the toggle by overriding the block `sw_custom_field_set_renderer_card_form_renderer`, but on 6.8 the mt-tabs migration (#17339) renders a new block `sw_custom_field_set_renderer_card_mt_tabs_form_renderer` that the extension does not override, so the toggle is dropped.

### Expected behaviour
Same as 6.7: each custom field shows a "Change: <field>" toggle on the left plus the field input on the right, so the field can be selected and bulk-applied.
Image

### How to reproduce
1. Enable 6.8 (`FEATURE_ALL=major`); in the admin console `Shopware.Feature.isActive('v6.8.0.0')` returns `true`.
2. Settings → Custom fields → create a set assigned to **Customers** with one field.
3. Customers → select 2+ customers → **Edit** (bulk edit) → scroll to **Custom fields** → open the set's tab.

Result: no "Change: <field>" toggle; only the bare field is shown.
Check: `document.querySelectorAll('.sw-bulk-edit__custom-fields .sw-bulk-edit-custom-fields__change').length` is `0` (while `.mt-tabs__item` is present).

Also reproducible in Product bulk edit. Blocks acceptance test `BulkEdit/BulkEditCustomerInformation.spec.ts` on major (#18168).

ngocblue:
## Why
On 6.8 (`v6.8.0.0`) the custom-field set tabs in bulk edit render as **mt-tabs**. The customer bulk-edit task still located the tab via `locator("a")` and scrolled it into view via the `.sw-tabs__arrow--right` button — neither exists in mt-tabs — so `BulkEditCustomerInformation.spec.ts` times out on the major/nightly-major acceptance run.

## What
`CustomerBulkEdit` page object + `BulkEditCustomers` task:
- Locate the set tab by text on `.mt-tabs__item` instead of `locator("a")` (the tab is now a `<button>`).
- Click the bulk-edit "Change" toggle by its class (`.sw-bulk-edit-custom-fields__change`) instead of `getByRole("checkbox")` (the mt-checkbox native input isn't directly actionable).
- Remove the obsolete `.sw-tabs__arrow--right` scroll loop and the `customFieldArrowRightButton` locator — Playwright auto-scrolls the tab into view on click.

Element-agnostic `getByText` keeps it working on the 6.7 render path too.

## Depends on / relates to
- Depends on product fix shopware/shopware#19236 — renders the "Change" toggle on the mt-tabs content block (without it there is no toggle to click).
- Unblocks shopware/shopware#18168 (stabilize `BulkEditCustomerInformation.spec.ts` on major).
- After merge + release, bump `@shopware-ag/acceptance-test-suite` in shopware/shopware `tests/acceptance/package.json`.

## Verification
- `tsgo` typecheck and `oxlint --type-aware` on the changed files: clean.
- Verified locally against a `FEATURE_ALL=major` instance carrying the #19236 product fix: the spec reaches and completes the custom-field step.


src/page-objects/administration/CustomerBulkEdit.ts
```diff
@@ -69,8 +68,7 @@ export class CustomerBulkEdit implements PageObject {
 
         //Custom fields
         const customFields = page.locator(".sw-bulk-edit__custom-fields");
-        this.customFieldArrowRightButton = customFields.locator(".sw-tabs__arrow--right");
```
frobel, line 72: Dangerous, because we want to be upwards compatible with 6.6.x, so it is not possible that way.

src/page-objects/administration/CustomerBulkEdit.ts
```diff
@@ -91,9 +89,7 @@ export class CustomerBulkEdit implements PageObject {
     }
 
     async getCustomFieldLinkByName(customFieldSetName: string): Promise<Locator> {
-        return this.page.locator("a").filter({
-            hasText: `${customFieldSetName}`,
-        });
+        return this.page.locator(".sw-bulk-edit__custom-fields .mt-tabs__item").getByText(customFieldSetName, { exact: true });
```
frobel: `.mt-tabs__item ` is not available in latest  6.6.x. So we can't add it that way. We have to make a version check here. OR?

frobel (changes requested):

src/page-objects/administration/CustomerBulkEdit.ts
```diff
@@ -91,9 +94,14 @@ export class CustomerBulkEdit implements PageObject {
...
 
     async getCustomFieldLinkByName(customFieldSetName: string): Promise<Locator> {
-        return this.page.locator("a").filter({
-            hasText: `${customFieldSetName}`,
-        });
+        // On V6_8_0_0 the custom-field-set tabs render as mt-tabs (a <button>, no scroll arrow);
+        // on 6.6.x/6.7 they are the legacy sw-tabs (<a>). Branch on the feature, not the version,
+        // because the nightly-major arm runs a 6.7.x build with the flag enabled.
+        if (this.instanceMeta.features["V6_8_0_0"]) {
+            return this.page.locator(".sw-bulk-edit__custom-fields .mt-tabs__item").getByText(customFieldSetName, { exact: true });
```
hienha-sw: Can we have string interpolation here:
e.g.
readonly customFieldSet: (name: string) => Locator;
...
this.customFieldSet = (name: string) => this.page.locator(".sw-bulk-edit__custom-fields .mt-tabs__item").getByText(name, { exact: true })

src/page-objects/administration/CustomerBulkEdit.ts
ngocblue: reverted to `getByRole("checkbox");` the class change was unnecessary — that timeout was the missing "Change" toggle ([shopware/shopware#19236](https://github.com/shopware/shopware/issues/19236), since fixed).

src/page-objects/administration/CustomerBulkEdit.ts
ngocblue: version-aware via` instanceMeta.features["V6_8_0_0"] `(FlowBuilder's pattern) — mt-tabs only on 6.8; legacy `<a>` + arrow loop untouched for 6.6.x/6.7.

src/page-objects/administration/CustomerBulkEdit.ts
ngocblue: converted to `customFieldSet: (name) => Locator` factory, dropped the needless **async**, kept the `V6_8_0_0` branch inside, removed the redundant ${name}.

